PR #11, opened 2019-08-15
Inflation Scheduling

techman83:
Scheduling Job Done!

What it contains
- Scheduler CLI + Docker Container
- Travis Build for Scheduler
- Fix branch switch logic noted by @HebaruSan 
- Clean up of dangerous defaults

There are more things we can do with the scheduling, but this is a like for like replacement of how we process inflations right now. The idea being that a job will be configured to launch this container on a schedule and if the queue is empty it'll file it and then exit.


docker-compose.yml
```diff
@@ -16,6 +16,21 @@ services:
...
+      dockerfile: scheduler/Dockerfile
+      context: .
+      target: dev
+    environment:
+      NETKAN_PATH: ${NETKAN_METADATA_PATH}
+      AWS_DEFAULT_REGION: ${CKAN_AWS_DEFAULT_REGION}
+      AWS_SECRET_ACCESS_KEY: ${CKAN_AWS_SECRET_ACCESS_KEY}
+      AWS_ACCESS_KEY_ID: ${CKAN_AWS_ACCESS_KEY_ID}
+      SQS_QUEUE: InboundDev.fifo
+      MAX_QUEUED: 1
```
HebaruSan: The commit history says this line was added for dev testing. Should it be updated or removed now?

techman83: You're likely missing a piece of information that only exists in my head right now. The docker compose is for development testing only. There are legit ways to use a compose file in prod, but we're already well tied to AWS and I know Elastic Container Service really well, so short of a provider bankrolling the infrastructure with some long term guarantees, I think it's a safe enough option for now (it's also essentially "free", as you only pay the ec2 costs).

HebaruSan: > The docker compose is for development testing only.

Ahh, OK. Never mind!

techman83:
Reducing filler code was a primary goal, I did a lot of unnecessary abstraction with the Perl code. I assumed I'd end up re-using a lot more and was pretty naive about the maintenance burden (was my first or second major dev project). We've been pretty feature starved due to how hard it is to add and change things, which was another motivator. I'm also pretty impressed how much we can get done with a relatively short list of external deps (the amount of times one of the 100+ deps in the Perl code base broke the build was a bit of a pain).

I have also realised that we don't need a second image for the scheduler. The 2 images are identical except for the entry point and I have some ideas how to make that more efficient. It came to me when I was thinking about a couple of other services that need to be done and realising we'll end up with so many images of essentially the same code!

I'll clean that up later today and merge it all in. Getting people like yourself + @DasSkelett across the code will be really great!

docker-compose.yml
techman83: Good to ask though! If something doesn't make sense, it's either locked in my brain or it properly doesn't make sense (the latter happens more than I'd like to admit :smile: )

PR #12, opened 2019-08-23
Webhooks + Status

techman83:
This Adds:
* NetKAN Status - allows for import/export of status for existing status page
* Container for re-use of existing webhooks

There needs to be some Docker Hub bits and pieces, that will come in a future PR.

### DIFF
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -47,7 +47,7 @@ def netkan():
 )
 @click.option('--key', envvar='SSH_KEY', required=True)
 def indexer(queue, metadata, token, repo, user, key,
-            status_db, debug, timeout):
+            debug, timeout):
     level = logging.DEBUG if debug else logging.INFO
     logging.basicConfig(
         format='[%(asctime)s] [%(levelname)-8s] %(message)s', level=level
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -1,12 +1,14 @@
 import click
-import logging
 import boto3
+import json
+import logging
 from pathlib import Path
 from .utils import init_repo, init_ssh
 from .github import GitHubPR
 from .indexer import MessageHandler
 from .scheduler import NetkanScheduler
 from .status import ModStatus
+
 
 @click.group()
 def netkan():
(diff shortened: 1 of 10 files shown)


netkan/netkan/cli.py
```diff
@@ -41,7 +50,8 @@ def netkan():
     help='Reduce message visibility timeout for testing',
 )
 @click.option('--key', envvar='SSH_KEY', required=True)
-def indexer(queue, metadata, token, repo, user, key, debug, timeout):
+def indexer(queue, metadata, token, repo, user, key,
+            status_db, debug, timeout):
```
HebaruSan: All of this file's functions now have a `status_db` parameter, but none of them seem to use it. Am I missing something, is there some way they get used implicitly?

netkan/netkan/cli.py
```diff
@@ -1,17 +1,26 @@
...
+
 
 @click.group()
 def netkan():
     pass
 
+
 @click.command()
+@click.option(
+    '--status-db', envvar='STATUS_DB',
```
HebaruSan: If I'm understanding `@click.option` correctly, this should be after the `--key` option to match up with the order of the parameters.

netkan/netkan/cli.py
```diff
@@ -113,5 +123,67 @@ def scheduler(queue, netkan, max_queued, debug):
...
+    help='DynamoDB Tablename'
+)
+@click.option(
+    '--status-bucket', envvar='STATUS_BUCKET', required=True,
+    help='Bucket to Dump status.json'
+)
+@click.option(
+    '--status-key', envvar='STATUS_KEY', default='status/netkan.json',
+    help='Bucket to Dump status.json'
```
HebaruSan: Copy-pasted help string needing an update

webhooks/run.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+cat > .ksp-cakn << END_CONFIG
```
HebaruSan: Probable typo in `.ksp-cakn` (a and k swapped)

netkan/netkan/cli.py
techman83: The order doesn't matter. These are passed through as `kwargs`, so aside from being neat it doesn't matter. There is a lot of duplication going on, but that's a  future thing that can be cleaned up.

netkan/netkan/cli.py
techman83: Yeah it could be dropped as passing through as an option won't work. The Status class is what needs it, but the click context isn't available until after. I made a comment regarding the table name. Probably worth dropping entirely as it really is just a dev thing.

netkan/netkan/cli.py
techman83: **whistles quietly**

webhooks/run.sh
techman83: **whistles quietly**

netkan/netkan/status.py
```diff
@@ -0,0 +1,103 @@
...
+)
+
+
+# The click context isn't available during the setup of this object and the
+# idea is that you'd specify a different table using a different class. Since
+# this is for dev and dev won't have access to the production table taking a
+# environment variable to override is acceptable for now.
+def table_name():
+    return os.getenv('STATUS_DB', 'NetKANStatus')
+
```
techman83: This is what I was referring to @HebaruSan

HebaruSan: Yup, I saw that. It was what prompted me to look more closely at the `--status_db` params, in fact.

netkan/netkan/cli.py
HebaruSan: This one (the `status_db` param of `indexer`) still seems to be there, while the others are removed.

techman83: Ahh late night coding. Glad I have a second set of eyes looking at this code!

PR #17, opened 2019-09-13
Download counting

### LINKED ISSUE #16: [Feature] Download Counts
So I entirely forgot that download counts were a thing. :grimacing: 

Suggestions
-----------
- This is currently part of the indexer, but needn't be
- A cli stub + class that generates this file and updates this when run
- A task definition + rule to run this daily

I'm happy enough to take care of getting the scheduling/definition logistics together. But it may be a little while before I have enough time to rewrite the total.

The requests library is already available in the netkan package and JSON is part of stdlib. JSON can be pretty printed/sorted the same way it is now
```
json.dumps({'4': 5, '6': 7}, sort_keys=True, indent=4))
```
If someone could hack a script together to produce the file, I'll do the logistics. Say make it under `netkan/netkan/download_counts.py` and I'll massage it it a class and click command. That'd be super appreciated!

HebaruSan:
## Motivation

In KSP-CKAN/NetKAN-bot#67 we started tracking download counts as reported by the various host servers in `CKAN-meta/download_counts.json`. These values are displayed in a column in GUI, which users can sort by to see the most/least downloaded mods.

Currently this isn't included in the Python rewrite.

## Changes

Now the download counting logic is ported from Perl to Python.

I'm in Windows at the moment and don't have a working Python3 environment (and I'm not sure how to run a `@click` command anyway), so I don't know whether this works, but the core logic should be there. Thanks in advance for reporting any bugs for me to fix.

Please feel free to add commits here. I attempted to set up a `@click` command based on the others, but I gave up on trying to guess the right syntax for docker and the other infrastructure.

- [ ] I forgot to make it add/commit/push in git. The Perl code piggybacks on the main push by the indexer, but now the download counter needs to handle git stuff itself.

Fixes #16.

### DIFF
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -4,11 +4,13 @@ import json
 import logging
 import time
 from pathlib import Path
+from datetime import timedelta
 from .utils import init_repo, init_ssh
 from .github import GitHubPR
 from .indexer import MessageHandler
 from .scheduler import NetkanScheduler
 from .status import ModStatus
+from .download_counter import DownloadCounter
 
 
 @click.group()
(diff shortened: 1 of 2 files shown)


techman83:
I'll close this and raise a new pr with the bits to finish this up shortly :slightly_smiling_face:

PR #18, opened 2019-09-13
Download Counts

techman83:
This builds on @HebaruSan 's previous PR and adds the necessary bits and pieces for the production service.

Also a bit of a clean up of the troposphere template, so I'm less likely to mess up the change over :sweat_smile:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -111,7 +111,7 @@ services:
       target: dev
     environment:
       SSH_KEY: ${CKAN_NETKAN_SSHKEY}
-      NETKAN_REPO: ${CKAN_METADATA_PATH}
+      NETKAN_REPO: ${NETKAN_METADATA_PATH}
       CKANMETA_REPO: ${CKAN_METADATA_PATH}
       GH_Token: ${CKAN_GH_Token}
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -105,3 +105,15 @@ services:
       - ./netkan:/home/netkan/netkan
       - ${HOME}/ckan_cache:/home/netkan/ckan_cache
     command: clean-cache --days 1
+  counts:
+    build:
+      context: netkan/.
+      target: dev
+    environment:
+      SSH_KEY: ${CKAN_NETKAN_SSHKEY}
+      NETKAN_REPO: ${CKAN_METADATA_PATH}
+      CKANMETA_REPO: ${CKAN_METADATA_PATH}
+      GH_Token: ${CKAN_GH_Token}
+    volumes:
+      - ./netkan:/home/netkan/netkan
+    command: download-counter
--- a/netkan/netkan/download_counter.py
+++ b/netkan/netkan/download_counter.py
@@ -1,85 +1,114 @@
-import logging
-import os
 import json
-import requests
+import logging
 import re
-from datetime import datetime
+import requests
 from pathlib import Path
+from json.decoder import JSONDecodeError
+
 
 class DownloadCounter:
 
-    def __init__(self, netkan_repo, ckan_meta_repo, github_token):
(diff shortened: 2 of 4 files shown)


techman83:
I did push it to the testing instance already and it [worked](https://github.com/techman83/pr_tester/commit/19b3daf448d0d7824a9e03c87dade3ff67448101)!

docker-compose.yml
```diff
@@ -105,3 +105,15 @@ services:
...
       - ${HOME}/ckan_cache:/home/netkan/ckan_cache
     command: clean-cache --days 1
+  counts:
+    build:
+      context: netkan/.
+      target: dev
+    environment:
+      SSH_KEY: ${CKAN_NETKAN_SSHKEY}
+      NETKAN_REPO: ${CKAN_METADATA_PATH}
+      CKANMETA_REPO: ${CKAN_METADATA_PATH}
```
HebaruSan: Shouldn't `NETKAN_REPO` and `CKANMETA_REPO` have different values?

netkan/netkan/download_counter.py
```diff
@@ -0,0 +1,114 @@
...
+            )
+            if count > 0:
+                self.counts[netkan.stem] = count
+
+    def write_json(self):
+        self.output_file.write_text(
+            json.dumps(self.counts, sort_keys=True, indent=4)
+        )
+
+    def counts_changed(self):
```
HebaruSan: I think this isn't called from anywhere. Looks like it should be used by `update_counts` to decide whether to call `commit_counts`?

docker-compose.yml
techman83: It's only for Dev and probably. But they can be the same, my testing repo has a single NetKAN in it. I messed up the commit logic after waiting for the update to finish successfully and didn't want to wait 😂 I'll fix it before I merge 🙂

netkan/netkan/download_counter.py
techman83: Oh gosh you're totally right, good catch! I was definitely getting tired when I finished this up.

GitPython will commit even when there are no changes, so you have to do those checks yourself.

I think the author intended it to be a low level API consumed by a project to put a friendly wrapper around it. Which comes with side effects of having to deal with the bits the git client normally takes care of for you.

I'll sort this out to, thank you!

HebaruSan:
Are we getting any errors in the new system? It's been a while since the counts were updated:

https://github.com/KSP-CKAN/CKAN-meta/commits/master/download_counts.json

techman83:
It looks like it is completing ok.
```
[2019-09-18 01:02:26,137] [INFO    ] Starting Download Count Calculation...
[2019-09-18 01:03:28,442] [ERROR   ] Failed to get count from https://raw.githubusercontent.com/InsaneDruid/Proton-M/master/ProtonMBreezeM.netkan
[2019-09-18 01:04:24,160] [ERROR   ] Failed to get count from https://raw.githubusercontent.com/jrossignol/KerbalSports/master/KerbalSports.netkan
[2019-09-18 01:09:59,126] [ERROR   ] Failed to get count from https://raw.githubusercontent.com/Poodmund/Outer-Planets-Mod/master/CKAN/OPM.netkan
[2019-09-18 01:16:18,735] [ERROR   ] Failed to get count from https://raw.githubusercontent.com/jrossignol/Strategia/master/Strategia.netkan
[2019-09-18 01:21:03,457] [INFO    ] Download counts match existing data.
[2019-09-18 01:21:03,458] [INFO    ] Download Counter completed!
```
It's possible the check for the existing data isn't right. Considering I wasn't even using it initially and was written when I was rather tired.

HebaruSan:
Yeah, it's almost impossible for the counts to be the same, all previous days had something like 1400+ lines changed in that file. Sounds like a condition is inverted.

PR #34, opened 2019-09-19
Clear old errors on successful inflation

HebaruSan:
## Problem

A day or two ago, SpaceDock had some downtime, and many modules hosted there were marked with errors. Now SpaceDock is fixed, but the errors are all still there:

http://status.ksp-ckan.space/

This means we can't tell whether a problem is still happening or already fixed.

## Cause

The new bot only sets the `last_error` attribute of the status object, and never clears it. Since the status objects persist long-term, this means the errors never go away.

## Changes

Now `last_error` is cleared upon success.

### DIFF
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -106,6 +106,7 @@ class CkanMessage:
         inflation_time = parse(self.CheckTime)
         attrs = {
             'success': self.Success,
+            'last_error': self.ErrorMessage,
             # We may wish to change the name in the inflator
             # as the index will set 'last_checked'
             'last_inflated': inflation_time,
@@ -117,8 +118,6 @@ class CkanMessage:
             attrs['ModIdentifier'] = self.ModIdentifier
         if self.indexed:
             attrs['last_indexed'] = datetime.now(timezone.utc)
-        if self.ErrorMessage:
-            attrs['last_error'] = self.ErrorMessage
         return attrs
 
     def _process_ckan(self):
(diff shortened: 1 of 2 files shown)


netkan/netkan/indexer.py
```diff
@@ -119,6 +119,8 @@ def status_attrs(self, new=False):
             attrs['last_indexed'] = datetime.now(timezone.utc)
         if self.ErrorMessage:
             attrs['last_error'] = self.ErrorMessage
+        else:
```
techman83: This is a little nit picky, but we can probably put this in our attrs block above with:
```python
'last_error': getattr(self, ErrorMessage, None)
```
Which is generally preferred to storing an empty string.

HebaruSan: Do we need `getattr`, or would `self.ErrorMessage` work? It does get set to `None` in `__init__`:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/5d7db8839f2951c6086af889ec6ae5c9a593565c/netkan/netkan/indexer.py#L17

techman83: Oh man. That's probably what I intended when I wrote this. Yeah, just self.ErrorMessage will do it. Good catch!

techman83: The status allows for this field to be set to `None`

https://github.com/KSP-CKAN/NetKAN-Infra/blob/5d7db8839f2951c6086af889ec6ae5c9a593565c/netkan/netkan/status.py#L33

HebaruSan: OK, force pushed that change (I don't like to keep commits that are simply errors).

techman83:
Nope, I agree. Excellent work!

PR #40, opened 2019-09-24
Skip scheduling netkans that only need webhooks

### LINKED ISSUE #10: [Feature] Skip web hook enabled modules in bot
This is an outgrowth of https://github.com/KSP-CKAN/CKAN/issues/2789#issuecomment-506253256. I assume that the initial functionality of the queue-based bot will be identical to NetKAN-bot, so ideas for further improvements should be submitted here for better tracking.

## Motivation

If a mod's `$kref` is on SpaceDock, our web hook will be notified of new uploads immediately, making the bot redundant some of the time. If a mod has a `$vref` property, then it's still necessary to crawl it, as its KSP-AVC version file may have a `"URL"` property pointing to a supplemental remote version file that could be changed at any time without notifying any web hook. But if there is no `$vref`, then the web hook should be sufficient by itself.

## Suggestion

If a module has a SpaceDock `$kref` and **no** `$vref`, it doesn't need to be added to the queue for inflation. Skip it.

There are 789 such modules, or about a third of the total, so this could cut the bot's total CPU load significantly.

HebaruSan:
## Motivation

There are 789 modules that are hosted on SpaceDock and have no `$vref`. These modules will be inflated by the SpaceDock webhook when they are uploaded, and the lack of a `$vref` means that their metadata is constant thereafter.

(A .version file's `URL` property points to a remote .version file which can be updated at any time to change the module's game version compatibility. This means that we always need to check modules with a `$vref`.)

## Changes

Now the scheduler checks for and skips such modules, so only modules that really need it will be inflated. This will reduce the amount of work the new bot needs to do on each pass, in exchange for having to parse the contents of the netkans into JSON.

Fixes #10.

### DIFF
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -101,14 +101,18 @@ def indexer(queue, metadata, token, repo, user, key,
     '--dev', is_flag=True, default=False,
     help='Disable AWS Credit Checks',
 )
-def scheduler(queue, netkan, max_queued, debug, dev):
+@click.option(
+    '--schedule-all', is_flag=True, default=False,
+    help='Schedule all modules even if we think they should be covered by webhooks'
+)
+def scheduler(queue, netkan, max_queued, debug, dev, schedule_all):
     level = logging.DEBUG if debug else logging.INFO
     logging.basicConfig(
         format='[%(asctime)s] [%(levelname)-8s] %(message)s', level=level
     )
     logging.info('Scheduler started at log level %s', level)
 
-    scheduler = NetkanScheduler(Path('/tmp/NetKAN'), queue)
+    scheduler = NetkanScheduler(Path('/tmp/NetKAN'), queue, force_all=schedule_all)
     if scheduler.can_schedule(max_queued, dev):
         init_repo(netkan, '/tmp/NetKAN')
         scheduler.schedule_all_netkans()
--- a/netkan/netkan/scheduler.py
+++ b/netkan/netkan/scheduler.py
@@ -2,14 +2,48 @@ import boto3
 import datetime
 import logging
 import requests
+import json
+import re
 from pathlib import Path
 from hashlib import md5
 
 
+class Netkan:
+
+    kref_pattern = re.compile('^#/ckan/([^/]+)/(.+)$')
+
+    def __init__(self, filename):
+        self.filename = Path(filename)
+        self.contents = self.filename.read_text()
+        self.json     = json.loads(self.contents)
+
+    def on_spacedock(self):
+        kref = self.json['$kref']
+        if not kref:
+            return False
+        (kind, mod_id) = Netkan.kref_pattern.match(kref).groups()
+        return kind == 'spacedock'
+
+    def has_vref(self):
+        return '$vref' in self.json
+
+    def hook_only(self):
+        return self.on_spacedock() and not self.has_vref()
+
+    def sqs_message(self):
+        return {
+            'Id':                    self.filename.stem,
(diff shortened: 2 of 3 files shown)


techman83:
This looks great from a glance. I'll take a more thorough look, probably in the morning.

We may want to consider having an option to 'schedule all', like most runs skip but once weekly do a full run or something?

HebaruSan:
Makes sense to me. Would that be a command line option?

techman83:
I'll merge this in, because I think it's an awesome enhancement. But I'll add some comments, that I'll likely address with the ideas I had for the webhooks.

netkan/netkan/scheduler.py
```diff
@@ -2,14 +2,48 @@
...
 
 
+class Netkan:
+
+    kref_pattern = re.compile('^#/ckan/([^/]+)/(.+)$')
+
+    def __init__(self, filename):
+        self.filename = Path(filename)
+        self.contents = self.filename.read_text()
+        self.json     = json.loads(self.contents)
```
techman83: I don't think there is a reason to store the contents of the file in memory, we can likely pass that straight to json.loads(). The `self.json` should really be called `raw` or `_raw` as it has been loaded in as a python object. The underscore is regularly used to identify it as private to the class, but it's just there as a 'YOYO' if you access this outside the class and expect it not to break.

netkan/netkan/scheduler.py
```diff
@@ -22,25 +56,17 @@ def __init__(self, path, queue, base='NetKAN/'):
...
-
-    def generate_netkan_message(self, filename):
-        content = Path(filename).read_text()
-        return {
-            'Id': filename.stem,
-            'MessageBody': content,
-            'MessageGroupId': '1',
-            'MessageDeduplicationId': md5(content.encode()).hexdigest()
-        }
+        return (Netkan(f) for f in self.path.glob('*.netkan'))
```
techman83: This is definitely a nit pick (we're only dealing with 1700 odd items here), but we're instantiating all these items into memory up front. Likely a little better to do it just before we need them.

netkan/netkan/cli.py
```diff
@@ -101,14 +101,18 @@ def indexer(queue, metadata, token, repo, user, key,
     '--dev', is_flag=True, default=False,
     help='Disable AWS Credit Checks',
 )
-def scheduler(queue, netkan, max_queued, debug, dev):
+@click.option(
+    '--schedule-all', is_flag=True, default=False,
+    help='Schedule all modules even if we think they should be covered by webhooks'
+)
```
techman83: Well it's more, web hooks can and do get missed. They are spray and pray by nature, So I think we should do full runs every now and then to ensure we don't miss stuff. How/When is definitely outside the scope of this PR.

netkan/netkan/scheduler.py
HebaruSan: Are we? I looked that up and it should be a generator with the parentheses instead of brackets.

HebaruSan: image

techman83: Well there you go! I missed that, I'll get another coffee ☕

netkan/netkan/scheduler.py
HebaruSan: `contents` is used below twice in `sqs_message`.

techman83: That's a point. Feels like a lot of duplication of data happening, in the scheme of things it probably doesn't matter.

PR #41, opened 2019-09-24
Use staging reason as PR body

### LINKED ISSUE #19: [Feature] NetKAN Staging Reason
The staging reason didn't make it across to the new code.

To rectify this, the Inflator will need to set an attribute containing the reason if there is one. Then the indexer can use `getattr(self, 'StagingReason', 'Default Description Here')` to set it.

HebaruSan:
If a `StagingReason` property is defined, it will be used as the body of the pull request for a staged module, otherwise the previous default string will be used.

Fixes #19.

### DIFF
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -153,11 +153,10 @@ class CkanMessage:
             self._process_ckan()
         if self.indexed:
             self.github_pr.create_pull_request(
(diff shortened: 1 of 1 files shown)


techman83:
f-strings are great and I'm not opposed to them in principal. Though they are new to 3.7, we have the luxury of controlling that in our container deployment so I'm just going to merge this. Also great use of the getattr :slightly_smiling_face:

PR #42, opened 2019-09-24
Populate release_date property

### LINKED ISSUE #6: [Feature] Generate release_date property in Indexer
This is an outgrowth of https://github.com/KSP-CKAN/CKAN/issues/2789#issuecomment-506136453. I assume that the initial functionality of the queue-based bot will be identical to NetKAN-bot, so ideas for further improvements should be submitted here for better tracking.

## Motivation

KSP-CKAN/CKAN#1155 requests a release date column in CKAN. Currently we do not have the ability to generate the metadata to populate such a column. It should correspond roughly to when a given version was indexed. `netkan.exe` can't generate this value by itself, because it requires knowing when the bot originally found a version, which can't be determined from checking the .netkan file or the .zip download. Some interaction with the CKAN-meta repo over time is necessary.

## Suggestion

The Indexer component could add and/or update this property between receipt of inflated metadata and saving it to disk. Sample code here:

https://gist.github.com/HebaruSan/665599c5c8fb9733c23c8aa64f29bc48

- If the file exists, pull the `release_date` property forward unchanged
- If the file exists but has no `release_date` property, set it to the file creation date in git
- If the file is brand new, set the `release_date` property to now

### Code notes

- `object_pairs_hook=collections.OrderedDict` preserves the ordering of the properties
- `.isoformat(timespec='seconds')` formats the datetime according to ISO8601, the format most commonly used by internet programmers and widely supported by languages and runtimes
- `git log --format='%aI' -- filename` prints the dates of all commits for a file in reverse chronological order, so the oldest is at the end of the list

Note that the `git log` command will only work with a full (non-shallow) clone of the repo, otherwise the full history of the file won't be present.

HebaruSan:
## Motivation

Users ask fairly frequently for the ability to sort modules by release date. Up to now we can't provide that info because the bot is stateless, and keeping track of the first time that we found a version requires state.

## Changes

Now the Indexer will add a `release_date` property to each module that it processes:

- If the file exists, it will pull the `release_date` property forward unchanged
- If the file exists but has no `release_date` property, it will set it to the file creation date in git
- If the file is brand new, it will set the `release_date` property to now

Note that the schema/spec currently do not have a `release_date` property. It will have to be added before this can be merged. In the meantime I want to see how many tests fail and why.

Fixes #6.
Fixes KSP-CKAN/CKAN#2916.

### DIFF
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -14,8 +14,7 @@ from .status import ModStatus
 class CkanMessage:
 
     def __init__(self, msg, ckan_meta, github_pr=None):
-        self.body = msg.body
-        self.ckan = Ckan(contents=self.body)
+        self.ckan = Ckan(contents=msg.body)
         self.ErrorMessage = None
         self.indexed = False
         for item in msg.message_attributes.items():
@@ -61,7 +60,7 @@ class CkanMessage:
     def write_metadata(self):
         self.mod_path.mkdir(exist_ok=True)
         with open(self.mod_file, mode='w') as file:
-            file.write(self.body)
+            file.write(self.ckan.body)
 
     def commit_metadata(self):
         index = self.ckan_meta.index
@@ -135,10 +134,19 @@ class CkanMessage:
             attrs['last_indexed'] = datetime.now(timezone.utc)
         return attrs
 
+    def get_release_date(self):
+        # Only look for a date if not already set
+        if not hasattr(self.ckan, 'release_date'):
+            self.ckan.find_release_date(self.mod_file, self.ckan_meta)
+            # Update coupled hash property
+            self.md5_of_body = hashlib.md5(self.ckan.body).hexdigest()
+
     def _process_ckan(self):
-        if self.Success and self.metadata_changed():
-            self.write_metadata()
-            self.commit_metadata()
+        if self.Success:
+            self.get_release_date()
+            if self.metadata_changed():
+                self.write_metadata()
+                self.commit_metadata()
         try:
             status = ModStatus.get(self.ModIdentifier)
             attrs = self.status_attrs()
--- a/netkan/netkan/metadata.py
+++ b/netkan/netkan/metadata.py
@@ -1,8 +1,9 @@
 import json
 import re
 from pathlib import Path
-from hashlib import sha1
+from hashlib import md5, sha1
 import uuid
+from collections import OrderedDict
 
 
 class Netkan:
@@ -15,7 +16,7 @@ class Netkan:
             self.contents = self.filename.read_text()
         else:
             self.contents = contents
-        self._raw = json.loads(self.contents)
+        self._raw = json.loads(self.contents, object_pairs_hook=OrderedDict)
         # Extract kref_src + kref_id from the kref
         if self.has_kref:
             self.kref_src, self.kref_id = self.KREF_PATTERN.match(self.kref).groups()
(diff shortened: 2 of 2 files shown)


techman83:
I have 2 comments without looking too closely at the code

- This is a break away from our long established single source of generation and letting the indexer make decisions
- The indexer should not be responsible for historical data

Now I'm not opposed to adding this to the meta data, but we really must flesh out the delineation of responsibility.

- We probably should have the discussion on the feature as it feels a little beyond the scope of an implementation detail.
- Historical data should be out of scope for the indexer. We have done this before with the hashes, we crawled all the metadata, added the hashes and mirrored them. In this case we can probably do it in a single pass as it is purely a metadata change, rather than downloading thousands of mods and calculating checksums.

HebaruSan:
> This is a break away from our long established single source of generation and letting the indexer make decisions

I think that's unavoidable. See the comments in KSP-CKAN/CKAN#2916; the Inflator doesn't know whether a file that it has generated already exists, so some other entity has to be in charge of copying forward already established release dates.

> The indexer should not be responsible for historical data

What's the reason behind that? The Indexer is built around a Git repo, so the historical data is all right there, and it is responsible for integrating new data into the existing database. Seems like a pretty natural fit.

techman83:
> I think that's unavoidable. See the comments in [KSP-CKAN/CKAN#2916](https://github.com/KSP-CKAN/CKAN/issues/2916); the Inflator doesn't know whether a file that it has generated already exists, so some other entity has to be in charge of copying forward already established release dates.

Yes, I think we're going to have to step away from this slightly (as. But I feel very strongly the source of this truth needs to be the inflator. It gives us scope to implement the release dates from the APIs we consume where available and leaves the indexer with the sole job of deciding of handling/publishing changes. As noted in KSP-CKAN/CKAN#2916

> What's the reason behind that? The Indexer is built around a Git repo, so the historical data is all right there, and it is responsible for integrating new data into the existing database. Seems like a pretty natural fit.

To clarify, I'm not opposed to creating a stub like I did for the restore_status. But it doesn't belong in the main indexing logic as it will be a once off task. The method itself looks ok to me, which we could use in a 'populate_historical_release_dates' if we wanted.

It also relies on a deep clone (which we could make optional), which we just don't need in regular operations. Part of my reasoning for using shallow cloning was to reduce our memory usage as during startup, as time goes on and as we add more services that'll become more important.

HebaruSan:
Ahh right, shallow clones. FWIW, I think I checked this before, but just now I got 137.7 MiB for a deep clone and 100.1 MiB for shallow, a difference of 25% of the size of the deep clone.

techman83:
> Ahh right, shallow clones. FWIW, I think I checked this before, but just now I got 137.7 MiB for a deep clone and 100.1 MiB for shallow, a difference of 25% of the size of the deep clone

That's space on disk, which in reality is neither here nor there. Memory usage during clone is where the bigger difference is.

HebaruSan:
Ahh, OK. Looks like that's on the order of a 45% difference:

```bash
$ /usr/bin/time --format='%M KiB' git clone https://github.com/KSP-CKAN/CKAN-meta.git
48804 KiB

$ /usr/bin/time --format='%M KiB' git clone --depth 1 https://github.com/KSP-CKAN/CKAN-meta.git
26616 KiB
```

HebaruSan:
Replaced by KSP-CKAN/CKAN#3059.

Issue #31, opened 2019-09-18
[Improvement] Refactor Download Counter


techman83:
Problem
-------
The download counter is a little awkward when it comes to making changes or catching places where an external api could result in the whole process failing.

Suggestions
-----------
- The data request should happen in one place. Log an error if it was unsuccessful, pass back a python object.
- There is a lot of function chaining, which makes it challenging to follow the logic

There's probably more, but made trying to fix spacedock disappearing while I was testing the fix for #29 a little beyond what I could before heading to my class after work!

techman83:
I have some ideas about how this might be achieved :-)

Issue #43, opened 2019-09-25
[Bug] Scheduler fails on 'KeyError' when no '$kref'


techman83:
We should use the dictionary get method here.
```python
Traceback (most recent call last):
  File ".local/bin/netkan", line 10, in <module>
    sys.exit(netkan())
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli.py", line 118, in scheduler
    scheduler.schedule_all_netkans()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/scheduler.py", line 80, in schedule_all_netkans
    for batch in self.sqs_batch_entries():
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/scheduler.py", line 65, in sqs_batch_entries
    if self.force_all or not netkan.hook_only():
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/scheduler.py", line 31, in hook_only
    return self.on_spacedock() and not self.has_vref()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/scheduler.py", line 21, in on_spacedock
    kref = self.json['$kref']
KeyError: '$kref'
```

techman83:
I'm going to look at this and a couple of other things shortly.

HebaruSan:
Since it took a little bit of effort, the list of netkans without krefs for posterity:

```
AdaptiveDockingNode.netkan
AIESAerospace-Unofficial.netkan
APModpack.netkan
AsteroidDay.netkan
AtmosphericSoundEnhancement.netkan
AUCAeronautictechnology.netkan
BananaForScale-large.netkan
BananaForScale-small.netkan
BearLandingLegLTB.netkan
[...]

PR #44, opened 2019-09-25
Fix Scheduler + Staging

techman83:
This fixes the unnecessary  staging branch creation and the small bug in the scheduler.

I also separated out the `Netkan` class as a bit of a preliminary work for #31

### DIFF
--- a/netkan/netkan/metadata.py
+++ b/netkan/netkan/metadata.py
@@ -1,0 +1,48 @@
+import json
+import re
+from pathlib import Path
+from hashlib import md5
+
+
+class Netkan:
+
+    kref_pattern = re.compile('^#/ckan/([^/]+)/(.+)$')
+
+    def __init__(self, filename):
+        self.filename = Path(filename)
+        self.contents = self.filename.read_text()
+        self._raw = json.loads(self.contents)
+
+    def __getattr__(self, name):
+        if name == 'kind':
+            kind, self.mod_id = self.kref_pattern.match(self.kref).groups()
+            return kind
+        if name == 'mod_id':
+            self.kind, mod_id = self.kref_pattern.match(self.kref).groups()
+            return mod_id
+        if name.startswith('on_'):
+            return self._on_kind(name.split('_')[1])
+        if name in ['kref', 'vref']:
+            if self._raw.get(f'${name}'):
+                return self._raw.get(f'${name}')
+        if name in self._raw:
+            return self._raw[name]
+        raise AttributeError
+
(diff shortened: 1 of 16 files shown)


netkan/netkan/metadata.py
```diff
@@ -0,0 +1,48 @@
...
+class Netkan:
+
+    kref_pattern = re.compile('^#/ckan/([^/]+)/(.+)$')
+
+    def __init__(self, filename):
+        self.filename = Path(filename)
+        self.contents = self.filename.read_text()
+        self._raw = json.loads(self.contents)
+
+    def __getattr__(self, name):
```
HebaruSan: This is going to make for difficult maintenance. If a developer needs to check what `on_spacedock` does to investigate a future bug, it's going to take a lot of reading and parsing and guessing.

Can we stick to explicitly naming functions with `def` the normal way, so they can be found?

techman83: This is a relatively common pattern in Python and means if we add a new kref type it's automatically supported by the class. The tests cover ensuring the variations work, though a comment or two probably wouldn't go astray.

I'm not 100% happy with the `__getattr__` block, but had to leave for dance class!

HebaruSan: I suppose the more serious problem is that it obscures the class's interface. If I have a reference to a `Netkan` object, the most important and frequent question I will ask is, "What can I do with this object"? With `def`'d functions and public variables assigned in the constructor, the answers are clear. With `__getattr__`, they could hardly be more murky. Imagine having to discover the existence of `on_curse` if you didn't already know exactly how this `__getattr__` implementation worked.

For "add a new kref type it's automatically supported by the class", I recommend making it an explicit parameter:

```python
    def is_on(kind):
        return kind == self.kref.kind # (or whatever implementation makes sense)
```

techman83: This as often a frustration with an undocumented class, but the reason behind this will become more apparent with the download counter. In big apps the maintenance burden is usually where the explicit patterns start becoming a burden as an interface change is more than a one liner.

I'll add some documentation when I get to work. I think the logic behind it will make more sense as I lean harder on the class.

techman83:
I think with #43 fixed, I'll close this and re-open including refactor for the Download Counter. It'll become a little more obvious where the strengths of this sort of pattern comes from.

It definitely was squashing #43 with a sledge hammer and belongs with a more thorough explanation of what is going on and why it's useful (I got carried away last night).

I'll cherry pick the staging fix and submit a PR for that shortly.

PR #46, opened 2019-09-27
Push Staging Branches only on 'Indexed'

techman83:
Currently the indexer indescrimently creates branches regardless of whether there are changes to submit. It's mostly just an annoyance right now, but it will make the branches available unmanageable at some point.

There is already an indexed flag set, so lets just use it! Also with the lack of testing going on in the webhooks container and it very much considered legacy, we probably only want to build that when we have a change to push (and not worth building automation for, lets just replace it!)

### DIFF
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -88,18 +88,28 @@ class CkanMessage:
             )
             branch.checkout()
         except GitCommandError:
-            branch = self.ckan_meta.create_head(self.mod_version)
+            if self.mod_version not in self.ckan_meta.heads:
+                branch = self.ckan_meta.create_head(self.mod_version)
+            else:
+                branch = getattr(
+                    self.ckan_meta.heads, self.mod_version
+                )
             branch.checkout()
-            self.ckan_meta.remotes.origin.push(
-                '{mod}:{mod}'.format(mod=self.mod_version)
-            )
         try:
             yield
         finally:
-            self.ckan_meta.remotes.origin.pull(
-                self.mod_version, strategy_option='ours'
-            )
-            self.ckan_meta.remotes.origin.push(self.mod_version)
+            if self.indexed:
+                # It's unlikely will hit a scenario where the metadata has
+                # changed upstream of us, but the bot should win if it does.
+                try:
+                    self.ckan_meta.remotes.origin.pull(
+                        self.mod_version, strategy_option='ours'
+                    )
+                except GitCommandError:
+                    pass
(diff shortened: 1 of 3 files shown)


netkan/netkan/indexer.py
```diff
@@ -88,18 +88,28 @@ def change_branch(self):
...
-                '{mod}:{mod}'.format(mod=self.mod_version)
-            )
         try:
             yield
         finally:
-            self.ckan_meta.remotes.origin.pull(
-                self.mod_version, strategy_option='ours'
-            )
-            self.ckan_meta.remotes.origin.push(self.mod_version)
+            if self.indexed:
```
HebaruSan: It looks like this won't work. `indexed` is set in `commit_metadata`, which is called in `_process_ckan`, **after** `change_branch`:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/805d9fc15a8016297c823980faf42126e12fe0ee/netkan/netkan/indexer.py#L64-L72

https://github.com/KSP-CKAN/NetKAN-Infra/blob/805d9fc15a8016297c823980faf42126e12fe0ee/netkan/netkan/indexer.py#L123-L137

https://github.com/KSP-CKAN/NetKAN-Infra/blob/805d9fc15a8016297c823980faf42126e12fe0ee/netkan/netkan/indexer.py#L151-L160

... so `self.indexed` will always be `False` in `change_branch`, and this block would never execute.

HebaruSan: Oh, that's what the `yield`/`with` thing is for, isn't it? That's kind of confusing.

Why not extend the existing `if self.indexed` block in `process_ckan`? Then the push and the creation of the pull request could be next to one another.

techman83: > Oh, that's what the `yield`/`with` thing is for, isn't it? That's kind of confusing.

That's another pythonism. It's a context manager, essentially the state inside the `with` gets cleaned up on the way out (you do have to write the cleanup process). In this instance, when in the staged context it'll change branches and on the way out, push any relevant changes and change back to master.
 
> Why not extend the existing `if self.indexed` block in `process_ckan`? Then the push and the creation of the pull request could be next to one another.

Basically to keep the commit process exactly the same for both staged and non staged ckans. It could go in the finally of the context manager, but my thinking is that if the PR raise blows up, I'd rather it do it after the branch has been changed back to master.

HebaruSan: OK, let's treat this as a "never mind".

PR #47, opened 2019-09-29
Refactor Download Counts

techman83:
The download counter was heavily inspired by the Perl version to get us across the line for production deployment of the new bot.

### Changes
- Move Netkan class into metadata file
- Extend Netkan class for counting downloads
- Tests!

Completed a full count pass with it [successfully](https://github.com/techman83/pr_tester/commit/0152c08e9a926d00d9b126d31b3f5b7515a9744c)

### DIFF
--- a/netkan/netkan/metadata.py
+++ b/netkan/netkan/metadata.py
@@ -1,0 +1,48 @@
+import json
+import re
+from pathlib import Path
+from hashlib import md5
+
+
+class Netkan:
+
+    kref_pattern = re.compile('^#/ckan/([^/]+)/(.+)$')
+
+    def __init__(self, filename):
+        self.filename = Path(filename)
+        self.contents = self.filename.read_text()
+        self._raw = json.loads(self.contents)
+
+    def __getattr__(self, name):
+        if name == 'kind':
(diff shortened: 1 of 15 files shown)


netkan/netkan/metadata.py
```diff
@@ -0,0 +1,65 @@
...
+        if filename:
+            self.filename = Path(filename)
+            self.contents = self.filename.read_text()
+        else:
+            self.contents = contents
+        self._raw = json.loads(self.contents)
+
+    def __getattr__(self, name):
+        # Extract kind + mod_id from the kref
+        if name in ['kind', 'mod_id']:
```
HebaruSan: So it turns out that `kind` is a valid defined metadata property outside of a kref context (it distinguishes packages from metapackages, and is used in the mirroring), so we'll need another name for that first part of the kref to avoid conflicting.

https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#kind

`Netkan.Model.RemoteRef` calls it `Source`. Maybe we could call it `kref_source` here?

techman83: Yeah, that's a point. I also don't like `mod_id` as it's kinda confusing. I'll refactor both.

PR #48, opened 2019-10-01
Flask webhooks

### LINKED ISSUE #8: [Feature] Provide a web hook for inflating a specific SpaceDock mod ID
This is an outgrowth of https://github.com/KSP-CKAN/CKAN/issues/2789#issuecomment-510297025. I assume that the initial functionality of the queue-based bot will be identical to NetKAN-bot, so ideas for further improvements should be submitted here for better tracking.

## Motivation

When a mod author uploads a new release to SpaceDock for a CKAN-indexed mod, SpaceDock uses the NetKAN-bot's web hooks to inflate that module immediately. This provides quicker indexing for such modules.

For this to work, someone needs to translate from a SpaceDock mod ID to a netkan module identifier. Surprisingly, this is currently done by SpaceDock (I searched for it in vain in our web hooks for a long time)! SpaceDock has an internal web app that searches a local copy of the NetKAN repo for modules matching a given `$kref`, which are then passed to the actual CKAN web hook:

https://github.com/KSP-SpaceDock/SpaceDock-Notify/blob/e7c2881d01948b5fc98e06de04df59af3763a23d/app.py#L31-L47

This architecture is a bit fishy. It means that if we make certain kinds of netkan-related changes to the spec (replace `$kref` with something else, add an optional suffix to the format for SpaceDock `$kref`s, add an alternate `$kref` format that works differently somehow, etc.), the SpaceDock web hook could stop working, and we could not fix it without collaborating with the SpaceDock team.

## Suggestion

Add a `/update/spacedock/<number>` web hook route that searches our own local copy of the NetKAN repo for a matching `$kref` and inflates it, essentially just like the above. This would better encapsulate the functionality, and it would allow SpaceDock to only worry about the ID numbers that it knows, while the CKAN team would maintain the mapping to identifiers.

### LINKED ISSUE #15: [Feature] Rewrite Webhooks
To get us across the line the existing webhooks have been used. We should switch this out for something like flask.

To Do
-----------
- [ ] Handle GitHub Webhooks
- [ ] Handle Spacedock Webhooks
- [ ] Handle Internet Archive Mirroring

We can probably sort #14 first and look at how we do this. We can likely have a TaskDefinition and a Webhooks Inbound queue or use a temporary Queue (Queues are free, traffic to them isn't, so I don't know the benefit for temporary for this use case, except it's faster and we don't care about having a few extra queues sitting idle). We can also use an archive queue and write a specific task definition for handling those tasks.

HebaruSan:
## Motivation

Currently the webhooks are handled by legacy Perl code. While this has worked surprisingly well, it's better to unify all functionality under the new Python framework.

## Changes

The `kspckan/netkan` image now allows an entrypoint that runs `gunicorn` listening on TCP/IP port 5000.

Required parameters (via `ENVIRONMENT_VARIABLES`):
- GitHub webhooks secret for authentication
- URL for NetKAN repo
- Name of SQS queue to which to send inflation requests; note that the AWS environment variables must also be set to access this queue

Routes handled:
- `/inflate` - Send an inflation request to the queue in response to a SpaceDock mod upload
- `/sd/inflate` - Send an inflation request to the queue in response to a SpaceDock mod upload, taking the SpaceDock ID number as the parameter instead of the CKAN identifier string
- `/gh/inflate` - Send an inflation request to the queue in response to a commit in the NetKAN repo
- `/gh/release` - Send an inflation request to the queue in response to a new release appearing in a mod's repo

Code-wise, this is done by adding a new `netkan/netkan/webhooks` directory with several new files:
- `__init__.py` - Imports the blueprint modules and exports a factory `create_app` function to generate the Flask app with the needed params
- `common.py` - Defines utility functions needed by `github_inflate.py` and `inflate.py`
- `github_inflate.py` - Defines blueprints to handle the GitHub webhooks that require inflation
- `github_utils.py` - Defines a `@signature_required` decorator that GitHub webhook handlers can use for authentication, the core of which was borrowed from some work @DasSkelett and others and I did recently for SpaceDock, which was used successfully on that project's test server
- `inflate.py` - Defines blueprint to handle the `/inflate` webhook
- `spacedock_inflate.py` - Defines blueprint to handle the `/sd/inflate` webhook

I did my best to update the infrastructure stuff, but I probably missed some things, so additional changes are likely needed. The code to generate the container for legacy webhooks is removed. The webhook portion of the `prod-stack.py` script is updated to share the common Python image and use the new `webhooks` command.

I tested by starting a server thusly (`sqs-run` sets the AWS env vars to the values we used for testing) (with the `@signature_required` decorator commented out so I didn't have to sign my test JSON):

```bash
python3 -m pip install --user -e netkan/.

sqs-run ~/.local/bin/netkan webhooks --secret asdf --netkan https://github.com/KSP-CKAN/NetKAN.git --ckan-meta https://github.com/KSP-CKAN/CKAN-meta.git --queue InboundDev.fifo --ia-access asdf --ia-secret asdf --ia-collection asdf
```

And then simulated clients in another shell:

```bash
wget http://127.0.0.1:5000/inflate --header='Content-Type: application/json' --post-data='{"identifiers":["Astrogator"]}' -O -

wget http://127.0.0.1:5000/gh/inflate --header='Content-Type: application/json' --post-data='{"commits":[{"added":["NetKAN/Astrogator.netkan"]}]}' -O -

wget http://127.0.0.1:5000/gh/mirror --header='Content-Type: application/json' --post-data='{"commits":[{"added":["Astrogator/Astrogator-v0.9.2.ckan"]}]}' -O -
```

So I was able to confirm that the web routes work, and that the queue submissions work, but I used fake credentials for archive.org and could not fully try out that piece (nor did I *want* to upload garbage data to archive.org).

Fixes #15. Fixes #8.


techman83:
I'll give it a thorough review after work, it's quite a large set of changes and will need some consideration before deploying. It however looks really solid from an initial skim.

netkan/netkan/webhooks/github_mirror.py
```diff
@@ -0,0 +1,315 @@
...
+        self.filename = Path(filename)
+        self.contents = self.filename.read_text()
+        self._raw     = json.loads(self.contents)
+
+    def can_mirror(self):
+        return (
+            self._raw.get('kind', 'package') == 'package'
+            and self._raw['download_content_type'] in Ckan.mime_to_extension
+            and self.redistributable()
+        )
```
techman83: True/False returns like this are difficult to parse and I don't know that they save anything compute wise. I think an if/else block that falls to return False or True (pending what's easier and makes sense) is clearer.

Also you can refer to class properties with `self`, ie `self.mime_to_extension`.

netkan/netkan/webhooks/github_mirror.py
```diff
@@ -0,0 +1,315 @@
...
+    def licenses(self):
+        lic = self._raw['license']
+        if isinstance(lic, list):
+            return lic
+        else:
+            return [lic]
+
+    def license_urls(self):
+        return [Ckan.license_urls[lic] for lic in self.licenses() if lic in Ckan.license_urls]
+
```
techman83: As above regarding `self` + class properties.

netkan/netkan/webhooks/github_mirror.py
```diff
@@ -0,0 +1,315 @@
...
+    def redistributable(self):
+        for lic in self.licenses():
+            if lic in Ckan.redistributable_licenses:
+                return True
+        return False
```
techman83: This doesn't match the current logic. We respect the most restrictive license within the package, unless all components are allowed to be redistributed freely, then we can't upload to the IA on the authors behalf.

netkan/netkan/webhooks/github_mirror.py
```diff
@@ -0,0 +1,315 @@
...
+    def download_size(self):
+        return self._raw['download_size']
+
+    def download(self):
+        return self._raw['download']
+
+    def mirrored(self, ia):
+        results = ia.search_items(self.mirror_item())
+        return True if results else False
+
```
techman83: I think we should have a similar pattern to the download counter here. A central metadata Class, extended for the Internet Archive.

We definitely need some tests for this.

netkan/netkan/webhooks/github_mirror.py
```diff
@@ -0,0 +1,315 @@
...
+                        'creator':     ckan.authors(),
+                        'collection':  current_app.config['ia_collection'],
+                        'subject':     'ksp; kerbal space program; mod',
+                        'mediatype':   'software',
+                        **({ 'licenseurl': lic_urls } if lic_urls else {}),
+                    }, {
+                        'Content-Type':           ckan.download_content_type(),
+                        'Content-Length':         ckan.download_size(),
+                        'x-amz-auto-make-bucket': 1,
+                    })
```
techman83: This is asking a lot of the flask app, one of the things that I found awkward with the previous architecture was that each piece was being asked to do too much, as we didn't have any other way to achieve it. 

I have some ideas here, but none of this work is wasted. It'll be easy enough to move it around as needed.

netkan/netkan/webhooks/github_mirror.py
HebaruSan: Hmm, that's a straight port of the current logic:

https://github.com/KSP-CKAN/NetKAN-bot/blob/4f218a2085f9a5e49f151c499a9192e97c5af728/lib/App/KSP_CKAN/Metadata/Ckan.pm#L202-L218

Remember that multiple licenses associated with a module do **not** indicate separate sub-licenses for different pieces, but rather multiple *options* for licensing for the overall package that the user can choose (that's why lists are so rarely used here):

https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#license

> A single license (v1.0) , or list of licenses (v1.8) may be provided. The following are both valid, the first describing a mod released under the BSD license, the second under the user's choice of BSD-2-clause or GPL-2.0 licenses.

So if there's a redistributable license in a list of multiple, that means that the entire package can be used under that license, not that some piece of it can be.

techman83: Yes, that's generally how licenses work; most permissive wins. However my understanding is that the metadata can't separate say source code under a permissive license, assets under a restricted license. So in the case of archiving we don't have the rights to redistribute unless the entire mod is covered by a permissive license. See KSP-CKAN/CKAN#22

I'm ok leaving it as the status quo, but technically I think my original logic is wrong.

HebaruSan: > the metadata can't separate say source code under a permissive license, assets under a restricted license

That's exactly right. A mod in that situation would list only the more restricted license in the `license` property, not both.

techman83: Oh good, the spec covers that! Well carry on :slightly_smiling_face:

netkan/netkan/webhooks/github_mirror.py
HebaruSan: Is that better? I kind of like having a reminder that a value is a static class property rather than an object property.

HebaruSan:
> I think we should separate out the Internet Archive work into a separate piece of work. There are a number of considerations that need to be thought about before replacing that piece and I reckon we can buy a lot of progress without getting hung up there.

I still think we should delay the whole PR together. It's all new code, and if we learn something important while polishing the archiving code, we can apply it to the other webhooks to ensure everything is the best it can be at the eventual go-live.

> I very much think we should be pushing mirroring requests to a queue and only deleting messages when the mirroring is successful.

That's a great idea. I'll look at defining the queue message format and splitting the code at that point (but I will probably still need help with the infrastructure pieces).

> Flask is also not designed to be deployed this way, the project suggests waitress, but I'm more familiar with gunicorn.

Oh, I assumed the `nginx` stuff would handle that. Isn't it the layer between the outside world and the app code?

netkan/netkan/webhooks/github_mirror.py
HebaruSan: Hmm, I have to admit that I like this style. It has simpler control flow (one statement with short-circuiting), and it's self-documenting, since the conditions that are needed are listed one after the next, without extra `return` statements or multiple `if` conditions. And if a new condition comes up it's easy to add it to the list. It gives some of the benefits of functional programming, in that it avoiding side effects and mutable variables in favor of a declarative approach.

HebaruSan:
OK, the mirroring now has its own request queue monitored by a new separate container. No tests yet, and there are probably other gaps, but I took care of as much as I could figure out this pass.

netkan/netkan/webhooks/github_mirror.py
techman83: I think being a Python project we should probably lean towards common patterns used by the language. Though I also don't work with functional languages that much and we don't have a style guide.

netkan/netkan/webhooks/github_mirror.py
techman83: Hmm, within Python it is unusual to refer to the class in this manner within the object itself. I don't know that it makes any difference in practice. A lot of python stuff is consistency related and while I don't blindly believe that everything needs to be "pythonic" there is value in it being familiar.

For now it's a nitpick, with good test coverage we'll be able to refactor as needed.

techman83:
> I still think we should delay the whole PR together. It's all new code, and if we learn something important while polishing the archiving code, we can apply it to the other webhooks to ensure everything is the best it can be at the eventual go-live.

I disagree. There is a lot of value in shaking down the deployment process and then building on it. It means we can really focus on getting the foundations right, rather then getting bogged down in trying to do it all at once.

I am aware the replacing the bot was a big bang approach, but we didn't have the foundations and it needed to be like for like. We have those now, so lets lean into them.
 
> That's a great idea. I'll look at defining the queue message format and splitting the code at that point (but I will probably still need help with the infrastructure pieces).

Absolutely lean on me for infrastructure assistance. It's what I do day in, day out.

> Oh, I assumed the `nginx` stuff would handle that. Isn't it the layer between the outside world and the app code?

It will be. But most frameworks have a built in server to ease development, where as things like gunicorn are designed to be a webserver and things like nginx being excellent connection handlers/managers (plus SSL management). I am more well versed with django, but I think this video might help you with  some of the reasons behind all this stuff if you are interested.

https://www.youtube.com/watch?v=oMHrDy62kgE

netkan/netkan/webhooks/github_mirror.py
HebaruSan: I'm pretty sure `return` statements and the `and` operator are common patterns used by Python. :wink: 

I understand that you feel protective towards this code since it's your baby, but I can only reassure you that other contributors also want only the best for it, and sometimes something that you're not familiar with can turn out to be a good thing. Can we table this till all the code is actually working?

techman83: Haha, that is true. That particular pattern is a touch foreign to me, but that's life and learning :slightly_smiling_face:, Long as we've got some test cases for it, we can always refactor if it becomes a maintenance burden. Though referring to class constants via the class name within the class, that I have a far stronger opinion about :wink: 

Absolutely, there are much bigger architectural decisions that are far more important. This was definitely a nit pick that got carried away. It's not that I'm super protective of the code per se, but I'm generally thinking about future me having to debug a production problem. Which from experience is never after an 8 hour sleep and just the right amount of coffee :smiley: 

As always, appreciate your effort and opinions!

netkan/netkan/webhooks/github_mirror.py
HebaruSan: `CkanMirror` now extends `Ckan`, with the IA functionality split out into the former, so I think that request is satisfied.

How do I run the tests?

HebaruSan:
Added some tests, but they don't cover anything that could actually fail, so...` ¯\_(ツ)_/¯`

netkan/netkan/metadata.py
```diff
@@ -63,3 +63,39 @@ def sqs_message(self):
...
+        if name in self._raw:
+            return self._raw[name]
+        if name == 'kind':
+            return self._raw.get('kind', 'package')
+        raise AttributeError
+
+    # These are properties, but they don't have @property because Python says:
+    #   TypeError: argument of type 'property' is not iterable
+    # ... which doesn't make sense because a property can still return an iterable type.
+
```
techman83: @property is a decorator, it allows a class function to appear as a property. There might be something in that, but if this works we can refactor later if it becomes annoying.

netkan/netkan/mirrorer.py
```diff
@@ -0,0 +1,284 @@
...
+            # If we can't mirror, then we're done with this message
+            logging.info(f'Ckan {ckan.mirror_item} cannot be mirrored')
+            return True
+        if ckan.mirrored(self.ia_session):
+            # If it's already mirrored, then we're done with this message
+            logging.info(f'Ckan {ckan.mirror_item} is already mirrored')
+            return True
+        with tempfile.NamedTemporaryFile() as tmp:
+            logging.info(f'Downloading {ckan.download}')
+            urllib.request.urlretrieve(ckan.download, tmp.name)
```
techman83: We can expose ckan_cache to the container to check there first, but I think this will be fine for now.

netkan/netkan/mirrorer.py
```diff
@@ -0,0 +1,284 @@
...
+        "MPL-1.1"           : 'https://www.mozilla.org/en-US/MPL/1.1/',
+        "Perl"              : 'http://dev.perl.org/licenses/',
+        "Python-2.0"        : 'https://www.python.org/download/releases/2.0/license/',
+        "QPL-1.0"           : 'https://opensource.org/licenses/QPL-1.0',
+        "W3C"               : 'https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document',
+        "Zlib"              : 'http://www.zlib.net/zlib_license.html',
+        "Zope"              : 'http://old.zope.org/Resources/License.1',
+        "Unlicense"         : 'https://unlicense.org/UNLICENSE',
+    }
+
```
techman83: These are class constants, we could use capitals or an underscore. It might even be worth defining it as a constant though this will be good for now as we're not using it anywhere else.

netkan/netkan/mirrorer.py
```diff
@@ -0,0 +1,284 @@
...
+            and self.redistributable
+        )
+
+    def mirrored(self, ia):
+        results = ia.search_items(self.mirror_item)
+        return True if results else False
+
+    def license_urls(self):
+        return [CkanMirror.license_url_dict[lic] for lic in self.licenses() if lic in CkanMirror.license_url_dict]
+
```
techman83: Nit Pick: It is unusual to reference a class constant via the class name within the class.

techman83:
> Added some tests, but they don't cover anything that could actually fail, so...` ¯\_(ツ)_/¯`

Indeed, those tests aren't for covering that, they're for covering refactoring and setting up a frame work for adding tests for when a bug is encountered.

netkan/netkan/mirrorer.py
HebaruSan: Well, the only way to improve common Python coding practice is one project at a time!

I want a self-documenting reminder that this is a class constant which is neither initialized per object nor a logical property of the object. The class name prefix gets me that, `self.` does not.

netkan/netkan/mirrorer.py
HebaruSan: Do you mean renaming it to `LICENSE_URL_DICT`? If so, I'm happy to do that. If not, please clarify how we would use capitals.

netkan/netkan/metadata.py
HebaruSan: On second thought, that comment was just venting and doesn't belong in the code long term. I'll remove it.

netkan/netkan/mirrorer.py
HebaruSan: Good idea, that should be simple enough to add...

netkan/netkan/mirrorer.py
HebaruSan: Apparent confirmation: https://stackoverflow.com/a/45702782/2422988
Will rename them accordingly...

HebaruSan:
After an extended wrestling session with `gunicorn`, I have a first pass towards using it:

- Remove `webhooks` command from `cli.py`
- Bring back the `kspckan/webhooks` docker image
  - But build it in the shared Python code folder
  - With `kspckan/netkan` as a base (not sure if this is a good idea, but it avoids having to figure out how much of the existing `Dockerfile` is necessary)
  - With a different `ENTRYPOINT` that calls `gunicorn`
- Replace function parameters of `webhooks.create_app` with environment variables

After building the image locally (with a temporary workaround to cover the mirroring queue not existing yet), I was able to access the `/inflate` hook after running it with this (`sqs-run` is my script to set the AWS variables):

```bash
sqs-run docker run \
	-e AWS_DEFAULT_REGION \
	-e AWS_ACCESS_KEY_ID \
	-e AWS_SECRET_ACCESS_KEY \
	--expose=5000 -p 5000:5000 \
	kspckan/webhooks \
	-e INFLATION_SQS_QUEUE=InboundDev.fifo \
	-e MIRROR_SQS_QUEUE=Mirroring.fifo \
	-e NETKAN_REPO='https://github.com/KSP-CKAN/NetKAN.git'
```

Fingers crossed that I'm on the right track!

netkan/netkan/webhooks/github_inflate.py
```diff
@@ -0,0 +1,56 @@
...
+
+# For release hook in module repo
+# Handles: https://netkan.ksp-ckan.space/gh/release?identifier=AwesomeMod
+# Putting this here instead of in a github_release.py file because it's small and quite similar to inflate
+@github_inflate.route('/release', methods=['POST'])
+@signature_required
+def release_hook():
+    id = request.args.get('identifier')
+    if not id:
+        return 'Param "indentifier" is required, e.g. http://netkan.ksp-ckan.space/gh/release?identifier=AwesomeMod', 400
```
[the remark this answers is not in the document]
HebaruSan: I need to proofread more carefully when copy-pasting :laughing: 

https://github.com/KSP-CKAN/NetKAN-bot/blob/4f218a2085f9a5e49f151c499a9192e97c5af728/lib/App/KSP_CKAN/WebHooks.pm#L82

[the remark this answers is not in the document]
DasSkelett: Oh wow, that misspelling held out a long time.

DasSkelett:
I think the docker-compose.yml needs an update too.

HebaruSan:
Is that better?

docker-compose.yml
```diff
@@ -53,18 +53,28 @@ services:
       - ${HOME}/ckan_cache:/home/netkan/ckan_cache
   webhooks:
     build:
-      context: webhooks/.
+      context: netkan/.
+      target: dev
```
DasSkelett: I think this `target` can be left out since the dockerfile doesn't use multi-stage builds.

HebaruSan: OK, I don't quite understand what that line does, but I'll take your word for this.

DasSkelett: If you look at netkan/Dockerfile, you see those FROM .. as ... lines.
These are for [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/).
So we first create a base image, and build different images for dev and production on it, where the dev image is "tagged" as `dev`...
https://github.com/KSP-CKAN/NetKAN-Infra/blob/19241c3c465d839d7e98f31f1fdfa77780cbc749/netkan/Dockerfile#L12-L19

...and the other one has no stage tag, which means it is the default one.
https://github.com/KSP-CKAN/NetKAN-Infra/blob/19241c3c465d839d7e98f31f1fdfa77780cbc749/netkan/Dockerfile#L21-L29

If you reference a Dockerfile in docker-compose to build from, without specifying a target the default one will be used.
So we need to specify `target: dev` to tell docker-compose to build the `dev` stage.

Hopefully this is somewhat understandable.

HebaruSan: Ohhhh!
So it just skips that final block that makes it print the help text when you start without parameters?
I thought it had something to do with the `dev` block later in this same file!

https://github.com/KSP-CKAN/NetKAN-Infra/blob/0e09894cb8b15af58d0d03c0ed24b1331e905802/docker-compose.yml#L130-L141

DasSkelett: Plus the final block doesn't run the `pip install --user -e netkan/.` command, but yes.
I'm not sure what the `dev` block of the compose is for either.

HebaruSan: What about `dev-stack.py`? I suppose that needs an update as well?

DasSkelett:
Not specifically a problem of this PR, but we have to clean up the environment variable naming and stuff regarding the metadata paths.
It would be nice to have a common naming scheme, and to have more descriptive names, plus using the same env var names inside and outside of the containers and in the code. I would suggest something along these lines:
```sh
CKAN_LOCAL_METADATA_PATH     # Path to local clone of KSP-CKAN/CKAN-meta
CKAN_UPSTREAM_METADATA_USER	 # The (GitHub) user/organization which the upstream metadata repo belongs to
CKAN_UPSTREAM_METADATA_REPO	 # The name of the Git(Hub) repo to push the generated metadata files to.
NETKAN_LOCAL_METADATA_PATH	 # Path to local clone of KSP-CKAN/NetKAN
```

There is a wild mix of the words "repo" and "path" currently, and whilst it might be technically correct that the directory under this path is also a git repository, mixing the terms creates a lot of confusion as to which is the upstream one, which is local and so on.

See `CKANMETA_REPO` and `METADATA_PATH` (inside Docker containers) for `CKAN_METADATA_PATH` (outside Docker containers)
or `NETKAN_REPO` and `NETKAN_PATH` (inside Docker containers) for `NETKAN_METADATA_PATH` (outside Docker containers).

HebaruSan:
@DasSkelett sounds like a job for https://github.com/KSP-CKAN/NetKAN-Infra/issues/new/choose :smile: 

But yes, I found some of the older names ambiguous or unclear (especially `METADATA_PATH`) and used new ones instead. What about using a `REMOTE` suffix to refer to the upstream URL/SSH strings instead of `PATH` or `REPO`? And I think the prefix part should match the repo name on GitHub, so `NETKAN` or `CKANMETA`.

Maybe some of this could be included in #49's proposed style guide; at least we could include a list of agreed upon "good" names and their meanings.

HebaruSan:
I decided to try addressing #8, and it was pretty easy so I added it to this PR. Once this is committed, the SpaceDock team will be able to update their notify URL to point to `https://netkan.ksp-ckan.space/sd/inflate` and retire their `SpaceDock-Notify` project.

netkan/netkan/mirrorer.py
```diff
@@ -0,0 +1,294 @@
...
+                self.ckan_meta_repo.remotes.origin.pull('master', strategy_option='ours')
+                # Start processing the messages
+                to_delete = []
+                for msg in messages:
+                    path = Path(self.ckan_meta_repo.working_dir, msg.body)
+                    if self.try_mirror(CkanMirror(path)):
+                        # Successfully handled -> OK to delete
+                        to_delete.append(deletion_msg(msg))
+                self.queue.delete_messages(Entries = to_delete)
+
```
techman83: You'll notice in the indexer we're only building an Indexer object every time we have messages to process, combined with a context manager. There was some method behind the madness, which was to avoid issues with file handles + memory within gitpython.

techman83:
I'm planning to review this properly over the weekend, it'll really good to retire the old webhooks and bring it all into align with the new world.

netkan/netkan/mirrorer.py
HebaruSan: Yeah, that was something I wanted to avoid because the core loop of the indexer is in `cli.py` instead of `indexer.py`. Does it actually solve something? Because in both cases one git repo object is initialized once at startup and then used and re-used as needed during processing. The indexer's core loop does the same thing as this new code: receive a batch of messages, process them, and delete them. I'm not seeing a functional run-time difference.

techman83: Yeah, I am conscious a lot is happening in the `cli.py`. I think we can refactor a lot of it by creating some click decorators and condensing all the common options. But didn't want to get bogged down with that initially.

https://gitpython.readthedocs.io/en/stable/intro.html#limitations

netkan/netkan/webhooks/spacedock_inflate.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+# For after-upload hook on SpaceDock
+# Handles: https://netkan.ksp-ckan.space/sd/inflate?mod_id=1234&event_type=update
+@spacedock_inflate.route('/inflate', methods=['POST'])
+def inflate_hook():
+    # Make sure our NetKAN repo is up to date
+    current_app.config['netkan_repo'].remotes.origin.pull('master', strategy_option='ours')
+    # Get the relevant netkans
+    nks = find_netkans(request.form.get('mod_id'))
```
DasSkelett: The comment in line 12 is wrong, isn't it?
You specified POST as request type, and later on use `request.form.get()` which is also for POSTs, but in the code comment the data is passed as URL query parameter for GETs.

HebaruSan: Good point, that comment was meant to be a succinct example but it does falsely imply a `GET` operation. Will update it.

netkan/netkan/mirrorer.py
```diff
@@ -0,0 +1,294 @@
...
+    def cache_find_file(self, ckan):
+        found = Mirrorer.CACHE_PATH.glob(f'**/{ckan.cache_prefix}*')
+        if found:
+            return found[0]
+        return None
+
+    def try_upload(self, ckan, file):
+        if ckan.download_hash.get('sha256') != hashlib.sha256(file.read()).hexdigest().upper():
+            # Bad download, try again later
```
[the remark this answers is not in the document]
HebaruSan: > As a very minor secondary, the get without a default will throw an AttributeError exception if there is no sha256.

Doesn't it return `None`? I thought that was the point of `get` existing as compared to the bracket operator.

netkan/Dockerfile.webhooks
```diff
@@ -0,0 +1,4 @@
+FROM kspckan/netkan
+RUN pip3 install --user flask gunicorn
+EXPOSE 5000
+ENTRYPOINT [".local/bin/gunicorn", "-b", "0.0.0.0:5000", "netkan.webhooks:create_app()"]
```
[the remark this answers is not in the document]
HebaruSan: Ah, I didn't know the entrypoint could be overridden without making a new image. That would definitely be preferable. Would that go in `prod-stack.py`?

netkan/netkan/mirrorer.py
```diff
@@ -0,0 +1,294 @@
...
+            with open(cached_file, mode='rb') as f:
+                logging.info(f'Found matching cache entry at {cached_file}')
+                return self.try_upload(ckan, f)
+        else:
+            # Download the file as needed
+            with tempfile.NamedTemporaryFile() as tmp:
+                logging.info(f'Downloading {ckan.download}')
+                urllib.request.urlretrieve(ckan.download, tmp.name)
+                logging.info(f'Downloaded {ckan.mirror_item} to {tmp.name}')
+                return self.try_upload(ckan, tmp.file)
```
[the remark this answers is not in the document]
HebaruSan: I believe `tmp.file` is a file handle:

https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

> tempfile.NamedTemporaryFile(mode='w+b', buffering=None, encoding=None, newline=None, suffix=None, prefix=None, dir=None, delete=True, *, errors=None)
> **The returned object is always a file-like object whose file attribute is the underlying true file object.**

netkan/netkan/mirrorer.py
```diff
@@ -0,0 +1,294 @@
...
+            return True
+        if ckan.mirrored(self.ia_session):
+            # If it's already mirrored, then we're done with this message
+            logging.info(f'Ckan {ckan.mirror_item} is already mirrored')
+            return True
+        cached_file = self.cache_find_file(ckan)
+        if cached_file:
+            # If the download is in the cache, use it
+            with open(cached_file, mode='rb') as f:
+                logging.info(f'Found matching cache entry at {cached_file}')
```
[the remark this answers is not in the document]
HebaruSan: I don't want `try_upload` to open the file because in its other usage, we already have an open file handle courtesy of `NamedTemporaryFile` (see below).

I'll switch this block to use `Path.open()` though.

netkan/netkan/mirrorer.py
HebaruSan: OK, I see now that `indexer.py` does do some extra cleanup in the form of `self.repo.close()`. Will add that here as well, since it appears to be safe (if undocumented):

https://github.com/gitpython-developers/GitPython/blob/6fd090293792884f5a0d05f69109da1c970c3cab/git/repo/base.py#L226-L239

netkan/netkan/mirrorer.py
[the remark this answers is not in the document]
HebaruSan: Will update the hash generation to avoid reading the whole file at once.

netkan/Dockerfile.webhooks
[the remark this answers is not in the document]
HebaruSan: Will make an attempt to override the entrypoint.

HebaruSan:
> Refactoring of names to go into a separate PR. I'd like us to be able to discuss + set naming conventions as a distinct thing.

Done in #58.

HebaruSan:
> Move Mirroring into a distinct PR.

I'm going to have to squash all the commits. It's too complicated to manage multiple commits for a change like this...

HebaruSan:
> Move Mirroring into a distinct PR.

Done in #59.

HebaruSan:
> Initially I think we should just get the new hooks running as a container on the production instance, that alone will shake out any deployment wrinkles. If we're happy, it's only a few lines to change which container nginx points at.

Is this something that requires me to make code changes? It sounds like a config thing. Please let me know.

techman83:
> I'm going to have to squash all the commits. It's too complicated to manage multiple commits for a change like this...

Totally understandable. The commits only matter while we're discussing changes (and that's purely a github UI thing).

> Is this something that requires me to make code changes? It sounds like a config thing. Please let me know.

Nope. Once we get #58 merged in we can rebase this. Any relevant changes to get it deployed I will make, push into this PR and merge.

Pretty excited to get this out, we've lost a lot of oversight with the old Webhooks. The logging really wasn't designed to go to the console, so half of its missing. It's working fortunately, but it'll get dicey if something doesn't work for whatever reason.

techman83:
I can get it to do things!

```bash
Attaching to netkan-infra_webhooks_1
webhooks_1     | [2019-10-23 10:22:18 +0000] [1] [INFO] Starting gunicorn 19.9.0
webhooks_1     | [2019-10-23 10:22:18 +0000] [1] [INFO] Listening at: http://0.0.0.0:5000 (1)
webhooks_1     | [2019-10-23 10:22:18 +0000] [1] [INFO] Using worker: sync
webhooks_1     | [2019-10-23 10:22:18 +0000] [8] [INFO] Booting worker with pid: 8
webhooks_1     | WARNING:netkan.webhooks:No signature provided in the request
webhooks_1     | WARNING:netkan.webhooks:X-Hub-Signature did not match the request data
webhooks_1     | 172.23.0.5 - - [23/Oct/2019:10:22:24 +0000] "POST /gh/inflate HTTP/1.0" 400 18 "-" "curl/7.58.0"
webhooks_1     | 172.23.0.5 - - [23/Oct/2019:10:22:35 +0000] "POST /inflate HTTP/1.0" 204 0 "-" "Wget/1.19.4 (linux-gnu)"
```
I'll add my changes required to get this running in production.

techman83:
Apologies for force pushing, I rebased on master to make sure we were all up to date. Once those tests pass I'll merge it in and deploy.

techman83:
That was a little bit of a mission, but it's deployed and working!

* Proxy
```
192.30.252.98 "POST /gh/inflate HTTP/1.1" 204 0 "GitHub-Hookshot/6760791"
192.30.252.98 "POST /gh/inflate HTTP/1.1" 204 0 "GitHub-Hookshot/6760791"
192.30.252.99 "POST /gh/mirror HTTP/1.1" 204 0 "GitHub-Hookshot/6760791"
```
* Legacy
```
[xKanHooks:11] info @2019-10-23 11:19:39> Mirroring: AllTweak/AllTweak-0.7.ckan in /usr/local/lib/perl5/site_perl/5.30.0/App/KSP_CKAN/WebHooks.pm l. 141
```
* New
```
[2019-10-23 11:15:21 +0000] [1] [INFO] Starting gunicorn 19.9.0
[2019-10-23 11:15:21 +0000] [1] [INFO] Listening at: http://0.0.0.0:5000 (1)
[2019-10-23 11:15:21 +0000] [1] [INFO] Using worker: sync
[2019-10-23 11:15:21 +0000] [9] [INFO] Booting worker with pid: 9
[23/Oct/2019:11:17:17 +0000] "POST /gh/inflate HTTP/1.0" 204 0 "-" "GitHub-Hookshot/6760791"
[23/Oct/2019:11:18:13 +0000] "POST /gh/inflate HTTP/1.0" 204 0 "-" "GitHub-Hookshot/6760791"
```

PR #56, opened 2019-10-20
Add VS Code Configuration

techman83:
## Motivation
Having a consistent environment for develop and testing is really helpful. VS Code has a lot of mind share and drive behind it, allowing us to bundle up a lot of our requirements and dependencies in an easy to use package.

## Changes
This PR adds a standard vs code environment, including plugin recommendations, development dependencies, and automatically opening up a container with the netkan project installed in editable mode.

For example, opening a terminal after the container has been built and launched
```bash
netkan@39991ca05cb3:~/workspace$ netkan --help
Usage: netkan [OPTIONS] COMMAND [ARGS]...

Options:
  --help  Show this message and exit.

Commands:
  clean-cache
  download-counter
  dump-status
  export-status-s3
  indexer
  redeploy-service
  restore-status
  scheduler
```

## Things of Note

- I've only tested this under Linux, the infrastructure is all Linux based and the project makes that assumption.
- VS Code uses a search for test discovery, instead of the internal unittest logic. The first time you run it, it will complain no tests are found. If you run the tests again, they work. I've yet to find a work around beyond renaming all the test files with 'test' in the names :roll_eyes: 
- We can write launch tasks for each of the different tasks, with their own environment variables. But I have left that as a future enhancement we can do.

### DIFF
--- a/netkan/Dockerfile
+++ b/netkan/Dockerfile
@@ -10,13 +10,13 @@ USER root
 RUN rm -Rf /netkan
 
 FROM python:3.7 as dev
-COPY --from=base /home/netkan /home/netkan
-RUN useradd -Ms /bin/bash netkan
-RUN chown -R netkan:netkan /home/netkan
-ADD run_dev.sh /usr/local/bin/
+RUN useradd -ms /bin/bash netkan
+ADD . /netkan
 WORKDIR /home/netkan
+ADD run_dev.sh /usr/local/bin/
+ENV PATH "$PATH:/home/netkan/.local/bin"
 USER netkan
-ENTRYPOINT ["/usr/local/bin/run_dev.sh"]
+RUN pip install --user /netkan/.[development]
 
 FROM python:3.7
 COPY --from=base /home/netkan /home/netkan
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,16 @@
+{
+	"name": "NetKAN Indexing",
+	"dockerComposeFile": [
+		"../docker-compose.vscode.yml"
+	],
+	"service": "netkan",
+	"workspaceFolder": "/home/netkan/workspace",
+	"settings": { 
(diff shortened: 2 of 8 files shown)


HebaruSan:
Well so far I'm impressed by how quickly VS Code launches. It's nearly instantaneous.

HebaruSan:
How do I get into that cool terminal thing? Pls EILI55.

netkan/Dockerfile
```diff
@@ -10,13 +10,13 @@ USER root
...
-RUN chown -R netkan:netkan /home/netkan
-ADD run_dev.sh /usr/local/bin/
+RUN useradd -ms /bin/bash netkan
+ADD . /netkan
 WORKDIR /home/netkan
+ADD run_dev.sh /usr/local/bin/
+ENV PATH "$PATH:/home/netkan/.local/bin"
 USER netkan
-ENTRYPOINT ["/usr/local/bin/run_dev.sh"]
+RUN pip install --user /netkan/.[development]
```
DasSkelett: I know I'm late to the party, but there's missing an entrypoint now, isn't it?
Why was `ENTRYPOINT [".local/bin/netkan"]` removed, or, shouldn't have `ENTRYPOINT [".local/bin/netkan"]` been added after pip?

techman83: Nah, this is the dev stage. So a combination of docker compose/vs code takes care of this.

DasSkelett: But in the normal docker-compose the entrypoint is not specified for most services, while targeting the dev stage:
https://github.com/KSP-CKAN/NetKAN-Infra/blob/a048da12cc36d5923314579aff0d83d33ae37040/docker-compose.yml#L7-L27

techman83: Good point. I haven't used the regular file since moving, the idea was to be able to spin up a whole dev set of services, but need to invest some more time making it all work in dev.

PR #58, opened 2019-10-20
Rename environment variables for consistency

HebaruSan:
## Problem

The new infrastructure uses environment variables and global Python variables in several places to pass along important parameters. 

- Inputs to `cli.py`
- Setting parameters in `prod-stack.py`
- Parameters for `docker-compose.yml`
- Inputs to legacy webhooks

Currently they aren't very consistent, and some of the names could be clearer:

| `cli.py` | `prod-stack.py` | Legacy webhooks | Comment |
| --- | --- | --- | --- |
| `METADATA_PATH`, `CKANMETA_REPO` | `CKAN_META` | `CKAN_meta` | This is the remote string for CKAN-meta. Not really a path, and somewhat ambiguous because the NetKAN repo also contains "metadata". |
| `NETKAN_PATH`, `NETKAN_REPO` | `NETKAN_HTTP` | `NetKAN` | This is the remote string for NetKAN. Not really a path, and it might not always use HTTP. |

## Changes

Now a consistent scheme is adopted:

- `[NETKAN|CKANMETA]_[REMOTE|USER|REPO]`

  The first part is the repository name in all caps and without punctuation.

  The second part is the aspect of the repo being referenced; `REMOTE` is used for consistency with the `git remote` command. `USER`  and `REPO` now exclusively refer to the pieces of the full GitHub repo URL (`KSP-CKAN` and `CKAN-meta`).

The existing variables are updated across the board to conform to this. This will make it easier to tell what a variable name means and what name to use when writing new code.

The following are **not** changed because I either don't know or can't access the other endpoint that sets or uses them:

- Legacy webhooks (we'll still set their old variables to avoid breaking them; the new webhooks in #48 will use the consistent scheme described above)
- `docker-compose.yml` (I don't know where the input variables are set, so they're left as-is)

This was split off from #48 (which will need to be re-based after this is merged).


netkan/netkan/cli.py
```diff
@@ -105,18 +105,18 @@ def indexer(queue, metadata, token, repo, user, key,
     '--schedule-all', is_flag=True, default=False,
     help='Schedule all modules even if we think they should be covered by webhooks'
 )
-def scheduler(queue, netkan, max_queued, debug, dev, schedule_all):
+def scheduler(queue, netkan_remote, max_queued, debug, dev, schedule_all):
     level = logging.DEBUG if debug else logging.INFO
```
techman83: Option is still `--netkan`, so the named param won't match

netkan/netkan/cli.py
```diff
@@ -105,18 +105,18 @@ def indexer(queue, metadata, token, repo, user, key,
...
     )
     logging.info('Scheduler started at log level %s', level)
 
-    scheduler = NetkanScheduler(Path('/tmp/NetKAN'), queue, force_all=schedule_all)
-    if scheduler.can_schedule(max_queued, dev):
-        init_repo(netkan, '/tmp/NetKAN')
-        scheduler.schedule_all_netkans()
-        logging.info("NetKANs submitted to {}".format(queue))
+    sched = NetkanScheduler(Path('/tmp/NetKAN'), queue, force_all=schedule_all)
+    if sched.can_schedule(max_queued, dev):
```
techman83: What was the motivation for shortening `schedule`?

netkan/netkan/cli.py
```diff
@@ -239,14 +237,14 @@ def clean_cache(days):
     '--debug', is_flag=True, default=False,
     help='Enable debug logging',
 )
-def download_counter(netkan, ckan_meta, token, key, debug):
+def download_counter(netkan_remote, ckan_meta, token, key, debug):
```
techman83: Same as previous param doesn't match option.

netkan/netkan/scheduler.py
```diff
@@ -72,29 +72,21 @@ def can_schedule(self, max_queued, dev=False):
             #       around 40 credits, with an accrue rate of 24/hr. So running every
             #       2 hours should see using just a touch less than we gain in that
             #       time period.
-            credits = 0
+            creds = 0
             try:
```
techman83: Same as previous comment regarding shortening of the variable.

netkan/netkan/cli.py
HebaruSan: Fixing this:

image

netkan/netkan/scheduler.py
HebaruSan: image

https://docs.python.org/2/library/constants.html#credits

netkan/netkan/cli.py
HebaruSan: Ahh, I thought these were matched up by position, and I didn't want to break any existing command line usages. But apparently we can add a new name without removing the old one:

https://click.palletsprojects.com/en/7.x/options/

HebaruSan: Fixed, now it works with either `--netkan` or `--netkan-remote`.

netkan/netkan/cli.py
HebaruSan: Fixed, now it works with either `--netkan` or `--netkan-remote`.

HebaruSan:
OK, I have removed the files that were only being changed for linting.

netkan/netkan/scheduler.py
techman83: Aha, didn't know credits was a constant, good catch. Though for reference, we don't care about python2, but it is a constant in 3 as well.

techman83:
This is awesome! Thanks for persisting, a narrow scope makes it much easier to review and get deployed (especially when it comes to infrastructure changes). I'll get this out the door after work.

netkan/netkan/cli.py
techman83: It's redefining the function name, which as it's the cli it's in the global (outer) scope. I'll wear that as a facepalm on my part :man_facepalming:

techman83:
Stack is updated and everything looks great!

PR #59, opened 2019-10-20
Mirroring webhook

HebaruSan:
## Motivation

See #48; we need to recreate the old webhooks in Python. The part that mirrors mods to archive.org is particularly complicated and so we're splitting it out into its own pull request.

## Changes

A new webhook is added to handle mirroring requests.
New environment variable parameters:
- Name of SQS queue to which to send mirroring requests; note that the AWS environment variables must also be set to access this queue

Routes handled:
- `/gh/mirror` - Send a mirroring request to the mirroring queue in response to a commit in the CKAN-meta repo

Code files:
- `github_mirror.py` - Defines a blueprint to handle the GitHub webhook that interacts with archive.org

A new `mirrorer` command is added to `cli.py` that listens to a queue of requests for mirroring and uploads the relevant files to archive.org. Params:

- Name of SQS queue to monitor for mirroring requests
- Visibility timeout for queue messages
- URL for CKAN-meta repo
- Access, secret, and collection strings for archive.org mirroring

This was split out from #48 and builds upon it. We will need to rebase this PR as that one changes.

### DIFF
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -17,6 +17,7 @@ from .download_counter import DownloadCounter
 from .ticket_closer import TicketCloser
 from .auto_freezer import AutoFreezer
 from .spacedock_adder import SpaceDockAdder
+from .mirrorer import Mirrorer
 
 
 @click.option(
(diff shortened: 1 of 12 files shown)


techman83:
I'll review this once we've got #58 + #48 sorted.

HebaruSan:
Rebased to resolve conflicts with #87.

HebaruSan:
Rebased to resolve conflicts with #96.

HebaruSan:
In latest commit, a new `netkan mirror-purge-epochs` command is created pursuant to recent discussion of KSP-CKAN/CKAN#2938. `Mirrorer.run` is renamed `process_queue` (and the queue-related parameters are migrated to it from the constructor) to make room for a sister `purge_epochs` driver function.

In contradistinction to previous commits, this time I created an archive.org account, retrieved my s3 credentials, and ran a model test script to make sure it works. I got 400 code responses to `modify_metadata` which I assume are due to my user not having permissions.

There was a `debug` parameter from before #86, now removed.
Also the class constants are now referenced through `self`.

netkan/tests/mirrorer.py
```diff
@@ -0,0 +1,60 @@
+import unittest
+
+from netkan.mirrorer import Mirrorer, CkanMirror
+
+
+# I don't know how to test Mirrorer, it needs a queue and archive.org access
+
```
techman83: This is indeed a challenge. I did build an entire replica of the API for testing NetKAN-bot, which is great but I don't think the value it adds matches the effort required.

Testing wise, we want to think about what is it that is important to test, so that we can change code and have it yell loudly if something changes unexpectedly. Both the important cases are covered, Redistributable + Restricted. But we might want one for handling mods with epochs to ensure our item/title is being produced by CkanMirror correctly.

netkan/tests/mirrorer.py
```diff
@@ -0,0 +1,60 @@
...
+            "version":      "1.0.0",
+            "ksp_version":  "1.7.3",
+            "author":       [ "techman83", "DasSkelett", "politas" ],
+            "license":      [ "CC-BY-NC-SA-4.0", "GPL-3.0", "MIT" ],
+            "download_content_type": "application/zip",
+            "download_hash": {
+                "sha1": "DF564E21929EA07C624F822E5C43B7D0A3B0DBDF"
+            }
+        }""")
+
```
techman83: The linter will probably complain about the extra spaces here and below :slightly_smiling_face:

prod-stack.py
```diff
@@ -770,6 +775,23 @@
             },
         ]
     },
+    {
+        'name': 'Mirrorer',
+        'command': 'mirrorer',
+        'memory': '156',
```
techman83: We can drop the memory and let the defaults handle it (we're getting tight on allocated memory for the service)

netkan/netkan/cli.py
```diff
@@ -309,12 +310,77 @@ def ticket_closer(token, days_limit):
...
+@click.option(
+    '--ia-collection', envvar='IA_collection',
+    help='Collection to put mirrored mods in on Internet Archive',
+    required=True,
+)
+def mirror_purge_epochs(ckan_meta, ia_access, ia_secret, ia_collection):
+    Mirrorer(
+        init_repo(ckan_meta, '/tmp/CKAN-meta'),
+        ia_access, ia_secret, ia_collection
+    ).purge_epochs()
```
techman83: I think we need a dry run option for this. It'd be nice to know what actions it's performing before committing to the action (being something we'll probably only run once).

netkan/tests/mirrorer.py
HebaruSan: Indeed, it complained about several things...

HebaruSan:
Before I push some fixes, a few concerns:

1. If merging the Adder with memory=156 caused problems, is it safe to merge another 96MB container now?
2. We should have explicit agreement about the future of epochs. **The current client includes epochs in its fallback URLs**, so it's not safe to remove them yet. The `mirror-purge-epochs` command is meant to be a tool we (eventually, somehow) use **in the future** to deal with epochs and mirroring, **not** run right away after this is merged.

netkan/netkan/cli.py
HebaruSan: Good idea, added this.

HebaruSan:
> builder policy + instance policy need access to the queue, along with a command to restart the service on build

Did my best to address this based on the fix PRs for the adder.

techman83:
> * f merging the Adder with memory=156 caused problems, is it safe to merge another 96MB container now?

That's a reasonable point, but I think I'm happy enough to merge this (probably tomorrow night after work), throw some tests at it and then drop the legacy hooks. Which will give us the room for the extra service. We may want to come up with a way to only spin these up as necessary, I have a refactor in mind that might help there. All the SQS listeners are almost the same, we could build a base class and ensure each one that inherits from it has a `process_message` or `process_batch` method that does all the hard work.

> * We should have explicit agreement about the future of epochs. **The current client includes epochs in its fallback URLs**, so it's not safe to remove them yet. The `mirror-purge-epochs` command is meant to be a tool we (eventually, somehow) use **in the future** to deal with epochs and mirroring, **not** run right away after this is merged.

I agree, lets stick with like for like right now. Then we can decide what we want to do from there next.

PR #65, opened 2019-10-23
Pull NetKAN before use in webhooks

HebaruSan:
## Problem

I think the current webhooks can pass outdated metadata to the Inflator. If we change a file in NetKAN, the webhooks don't update their local repo before reading the files, so the version of the file from the last container refresh would be used.

(The exception is the new (and so far unused) `/sd/inflate` hook, which pulls the master branch before using it. I must have been thinking more carefully about new code as compared to the other hooks which were ports of the legacy Perl code.)

## Changes

Now we pull the master branch before we use files from the NetKAN repo.

The `pylint` comments are to inform pylint that it is OK that the blueprint variables are in lowercase.

### DIFF
--- a/netkan/netkan/webhooks/github_inflate.py
+++ b/netkan/netkan/webhooks/github_inflate.py
@@ -5,7 +5,7 @@ from .common import netkans, sqs_batch_entries
 from .github_utils import signature_required
 
 
-github_inflate = Blueprint('github_inflate', __name__)
+github_inflate = Blueprint('github_inflate', __name__) # pylint: disable=invalid-name
 
 
 # For after-commit hook in NetKAN repo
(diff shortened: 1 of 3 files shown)


netkan/netkan/webhooks/github_inflate.py
```diff
@@ -49,6 +49,8 @@ def ids_from_commits(commits):
 
 
 def inflate(ids):
+    # Make sure our NetKAN repo is up to date
+    current_app.config['netkan_repo'].remotes.origin.pull('master', strategy_option='ours')
```
techman83: In this instance, we probably want our remote to win. The indexer sets 'ours', because it's the source of the truth. The webhooks/schedulers are consumers of the data.

PR #67, opened 2019-10-24
Don't try to extract kref pieces if there's no kref

### LINKED ISSUE #66: [Bug] Bot only crawling about 30 mods at the moment
## Problem

1. Go to http://status.ksp-ckan.space/
2. Sort by Last Checked
3. Scroll down to where the times jump from short to long; currently it's here:
  image

The mods with recent times only number about 30. The rest aren't being checked.
Are the remaining mods not being scheduled? Is the Inflator stopping mid-pass?

### Clues

- Started happening after #48 was merged
- The inflation webhooks are working
- Inflator hasn't been changed
- Indexer hasn't been changed
- Scheduler had a parameter and a local variable renamed, which should have had no effect, but who knows
- The most recently checked mod is pretty consistently `ContractConfigurator-KFiles`

HebaruSan:
## Problem

See #66; the bot is doing incomplete passes at the moment.

## Cause

This line from #48 throws when loading a netkan without a `$kref`:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/7b72e54ee985b7cc3dead5b2381b1a6eada021f5/netkan/netkan/metadata.py#L18-L20

Once the Scheduler encounters such a netkan (after about 30 mods), it throws this exception and stops:

```
  File "<stdin>", line 1, in <module>
  File "NetKAN-Infra/netkan/netkan/scheduler.py", line 26, in <genexpr>
    return (Netkan(f) for f in self.path.glob('*.netkan'))
  File "NetKAN-Infra/netkan/netkan/metadata.py", line 20, in __init__
    self.kref).groups()
  File "NetKAN-Infra/netkan/netkan/metadata.py", line 37, in __getattr__
    raise AttributeError
```

## Changes

Now we check `self.has_kref` first. If it's false, we set `kref_src` and `kref_id` to `None`.

Tests are added to cover the missing-`$kref` case.

Fixes #66.

### DIFF
--- a/netkan/netkan/metadata.py
+++ b/netkan/netkan/metadata.py
@@ -16,8 +16,11 @@ class Netkan:
             self.contents = contents
         self._raw = json.loads(self.contents)
         # Extract kref_src + kref_id from the kref
-        self.kref_src, self.kref_id = self.KREF_PATTERN.match(
-            self.kref).groups()
+        if self.has_kref:
+            self.kref_src, self.kref_id = self.KREF_PATTERN.match(self.kref).groups()
+        else:
+            self.kref_src = None
+            self.kref_id = None
 
     def __getattr__(self, name):
         # Return kref host, ie `self.on_spacedock`. Current krefs include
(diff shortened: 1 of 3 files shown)


techman83:
These are great! Thanks for adding a test case.

PR #68, opened 2019-10-25
Discord notifications

### LINKED ISSUE #52: [Feature] Send bot errors to Discord
CKAN's Discord has channels for notifications from GitHub and Travis. It would be awesome if errors from the new Infra code could be sent to a new channel there as well. Ideally this would include both inflation errors for specific modules and exceptions thrown by other components.

~~There's a `discord.py` module that could probably be used to do this.~~
Apparently that module targets a more interactive/reactive use case (and has great difficulty with simply dumping messages):
https://github.com/Rapptz/discord.py/issues/2391#issuecomment-546164552

HebaruSan:
## Motivation

See #52, we'd like to see error messages in Discord. It provides several advantages:

- Ease of listing and reviewing recent errors
- Instant notifications
- Convenient discussion of how to address them

## Background

Discord supports two methods of programmatic interaction:

- "Bots" (?) that can receive messages and react to them
- "Webhooks" that can send messages to a given channel on demand
  **NOTE**: Not to be confused with GitHub webhooks, which are not related in any way!

Both methods are provided by the `discord.py` module, but "Webhooks" are much less prominent in the documentation. They're also much easier to use (no `async`/`await`) and are a much closer fit for our requirements.

- https://discordpy.readthedocs.io/en/async/api.html
- https://techwithtim.net/tutorials/discord-py/setup/
- https://hackaday.com/2018/02/15/creating-a-discord-webhook-in-python/

## Changes

Now `notifications.py` contains:

- A `DiscordLogHandler` class that can be used to listen to `logging` calls and send them to Discord
- A `setup_log_handler` function that installs the above class for `error` level and higher messages if environment variables `DISCORD_WEBHOOK_ID` and `DISCORD_WEBHOOK_TOKEN` are defined

Now `cli.py` and `webhooks.create_app` call `setup_log_handler` to enable Discord logging for the existing toolkit.

Now the Indexer calls `logging.error` when inflation fails for a module with a new error message. This would allow it to appear on Discord, notifying developers immediately when there's a new problem. Note that some inflation errors include text that changes every time (temp file paths), so we may want to clean those up.

`setup.py` is updated to install the `discord` module.

Fixes #52.

## Setup

I did these parts already:

- [X] Create a `#netkan-bot` channel in the CKAN Discord
- [X] Add a Discord webhook in the channel

To be done at or after merge:

- [x] Move the `#netkan-bot` channel into the `NOTIFICATIONS` category (I don't have privileges to do this)
- [ ] Provide the `DISCORD_WEBHOOK_ID` and `DISCORD_WEBHOOK_TOKEN` environment variables to the containers

I have the ID and token for the Discord webhook I created and will share them privately as needed.

### DIFF
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -1,11 +1,11 @@
 import hashlib
+import logging
 from pathlib import Path, PurePath
 from collections import deque
-from dateutil.parser import parse
 from datetime import datetime, timezone
 from contextlib import contextmanager
+from dateutil.parser import parse
 from git import GitCommandError
-import logging
 
 from .status import ModStatus
 
@@ -46,8 +46,8 @@ class CkanMessage:
         return self.mod_file.stem
 
     def mod_file_md5(self):
-        with open(self.mod_file, mode='rb') as f:
-            return hashlib.md5(f.read()).hexdigest()
+        with open(self.mod_file, mode='rb') as file:
+            return hashlib.md5(file.read()).hexdigest()
 
     def metadata_changed(self):
         if not self.mod_file.exists():
@@ -58,8 +58,8 @@ class CkanMessage:
 
     def write_metadata(self):
         self.mod_path.mkdir(exist_ok=True)
-        with open(self.mod_file, mode='w') as f:
-            f.write(self.body)
+        with open(self.mod_file, mode='w') as file:
+            file.write(self.body)
 
     def commit_metadata(self):
         index = self.ckan_meta.index
@@ -67,7 +67,7 @@ class CkanMessage:
         commit = index.commit(
             'NetKAN generated mods - {}'.format(self.mod_version)
         )
-        logging.info('Committing {}'.format(self.mod_version))
+        logging.info('Committing %s', self.mod_version)
         self.indexed = True
         return commit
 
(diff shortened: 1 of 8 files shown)


netkan/netkan/indexer.py
```diff
@@ -137,10 +137,13 @@ def _process_ckan(self):
         try:
             status = ModStatus.get(self.ModIdentifier)
             attrs = self.status_attrs()
+            if not self.Success and status.last_error != attrs.last_error:
+                logging.error('New inflation error for %s: %s',
+                              self.ModIdentifier, attrs.last_error)
```
DasSkelett: This is potentially spammy, because error messages can include the name of the temporary download file (e.g. invalid ZIP files), and that one changes every run.
Currently this is only `TrimIndicator` though.

HebaruSan: Yeah, I am leaning towards removing the filename from that error. Can you think of a reason to keep it?

DasSkelett: Well, it helps finding the file for investigation purposes, but if you debug-log the file name, it should be enough.

HebaruSan: Fixed in KSP-CKAN/CKAN#2903.

prod-stack.py
```diff
@@ -698,6 +705,7 @@
                 ],
                 'secrets': [
                     'XKAN_GHSECRET',
+                    'DISCORD_WEBHOOK_ID', 'DISCORD_WEBHOOK_TOKEN',
                 ],
```
techman83: This doesn't need adjustment for me to merge, but we could add these as default secrets. Something like:

```python
secrets = [
    'DISCORD_WEBHOOK_ID', 'DISCORD_WEBHOOK_TOKEN',
    *container.get('secrets', [])
]
```
One less thing to think about when adding containers or services.

HebaruSan:
> if you can get me the tokens (or give me the power to generate them), I'll get the secrets sorted

Sent via PM on Discord.

PR #70, opened 2019-10-28
Copy resources to status objects

HebaruSan:
## Motivation

In KSP-CKAN/NetKAN-status#9 the status page is gaining the ability to show additional links for mods by extracting them from the status objects. The idea is to save steps in getting to all the various pages that are relevant to investigating a mod.

## Changes

This pull request copies the `resources` object's properties from an inflated module into the status object. Note that the `resources` property will be unavailable if inflation fails, but values saved from previous successful passes will still be there to aid in investigation.

### DIFF
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -1,3 +1,4 @@
+import json
 import hashlib
 import logging
 from pathlib import Path, PurePath
@@ -29,6 +30,7 @@ class CkanMessage:
         self.receipt_handle = msg.receipt_handle
         self.ckan_meta = ckan_meta
         self.github_pr = github_pr
+        self._raw = json.loads(self.body)
 
     def __str__(self):
         return '{}: {}'.format(self.ModIdentifier, self.CheckTime)
@@ -123,6 +125,8 @@ class CkanMessage:
             # If we have perfomed an inflation, we certainly
             # have checked the mod!
             'last_checked': inflation_time,
+            # Copy the links to the status page
+            **(self._raw.get('resources', {})),
         }
         if new:
             attrs['ModIdentifier'] = self.ModIdentifier
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -126,9 +126,10 @@ class CkanMessage:
             # If we have perfomed an inflation, we certainly
             # have checked the mod!
             'last_checked': inflation_time,
-            # Copy the links to the status page
-            **(getattr(self.ckan, 'resources', {})),
         }
+        resources = getattr(self.ckan, 'resources', None)
+        if resources:
+            attrs.resources = resources
         if new:
             attrs['ModIdentifier'] = self.ModIdentifier
         if self.indexed:
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -129,7 +129,7 @@ class CkanMessage:
         }
         resources = getattr(self.ckan, 'resources', None)
         if resources:
-            attrs.resources = resources
+            attrs['resources'] = resources
         if new:
             attrs['ModIdentifier'] = self.ModIdentifier
         if self.indexed:
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -8,6 +8,7 @@ from contextlib import contextmanager
 from dateutil.parser import parse
 from git import GitCommandError
 
+from .metadata import Ckan
 from .status import ModStatus
 
 
@@ -15,6 +16,7 @@ class CkanMessage:
 
     def __init__(self, msg, ckan_meta, github_pr=None):
         self.body = msg.body
+        self.ckan = Ckan(contents=self.body)
         self.ErrorMessage = None
         self.indexed = False
         for item in msg.message_attributes.items():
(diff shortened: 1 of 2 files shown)


netkan/netkan/status.py
```diff
@@ -36,6 +36,12 @@ class Meta:
     last_inflated = UTCDateTimeAttribute(null=True)
     success = BooleanAttribute()
 
+    homepage = UnicodeAttribute(null=True)
+    spacedock = UnicodeAttribute(null=True)
+    repository = UnicodeAttribute(null=True)
+    curse = UnicodeAttribute(null=True)
+    bugtracker = UnicodeAttribute(null=True)
+
```
techman83: Adding a bunch of specifically named fields isn't ideal, we either want to store a JSON payload (each field can hold up to 400Kb) or have a generic + type. I'm leaning towards a `resources` field, means we don't need to add a PR everytime we want to add some data (we're not doing any sort of filtering on this here). Though we should be aware that if our export grows beyond 10,240Kb we'll need to think about what we need to do (we're at like 400Kb, so we shouldn't have the problem for a while and I'd hope we've got everything polished enough to build an API).

netkan/netkan/indexer.py
```diff
@@ -123,6 +125,8 @@ def status_attrs(self, new=False):
             # If we have perfomed an inflation, we certainly
             # have checked the mod!
             'last_checked': inflation_time,
+            # Copy the links to the status page
+            **(self._raw.get('resources', {})),
         }
```
techman83: We should break out the CKAN metadata stuff from the mirroring PR, so we can consume it here.

netkan/netkan/status.py
HebaruSan: I'm very open to changing this, but I don't quite follow what you're suggesting. What would a generic + type look like?

netkan/netkan/indexer.py
HebaruSan: If you mean the `Ckan` class in `metadata.py`, that was part of the main webhooks PR and already available. I'll take a look at using it here...

HebaruSan: ... done. I tested analogous expressions as much as I could in a `python3` prompt, so hopefully this won't throw exceptions.

HebaruSan:
> Our status dumper may not be aware of the new fields

As far as I can tell, the status dumper will be OK:

- `export_to_s3` calls `export_all_mods` to generate the body
- `export_all_mods` returns a `dict` with mod identifiers as keys and `mod_attrs` as values
- `mod_attrs` converts its `self` into a `dict` by looping over all attributes and copying them, with special processing for `ModIdentifier` and `datetime`s

So the new fields would be copied in `mod_attrs`, just like the old ones.

netkan/netkan/status.py
techman83: We can use a [map type](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes). If we wanted to, we could just put the resources directly in there as they appear in the metadata. Which in future if we wanted to map this data in an API directly, we could retrieve it from different places. I'm actually leaning towards this option, because it matches the metadata as is (even though from a data structure layout it doesn't scale well).

In regards to mapping, I mean defining types (don't take these names as gospel)

```json
{
    "resources": {
        "provider": "github",
        "provider_uri": "https://github.com/foo/bar",
        "homepage": "https://foo.com/bar"
    }
}
```

Means that adding a provider doesn't involve spec changes, reduces code changes etc etc. But as noted, I think a resources map datatype and just put the relevant fields in there. If we want to change how we do resources in the future, doing it at the indexing end doesn't make sense. We're consuming the spec, not deciding it.

netkan/netkan/indexer.py
techman83: Now, do we want to breakout the metadata.py into a separate pr or merge the mirroring untested/unused for later inspection. The former would be easier for me to test + review, the later would get us moving on this faster.

HebaruSan: Like I said, `metadata.py` was already merged in #48 and has no changes in this PR.

techman83: Oh gosh, I totally mis-read that and had made the assumption it wasn't available as you hadn't used it. Too many interruptions this morning while I was absorbing the first coffee of the day!

netkan/netkan/status.py
HebaruSan: OK, replaced the discrete properties with one `resources` `MapAttribute`.

netkan/netkan/indexer.py
```diff
@@ -129,7 +129,7 @@ def status_attrs(self, new=False):
...
         resources = getattr(self.ckan, 'resources', None)
         if resources:
-            attrs.resources = resources
+            attrs['resources'] = resources
         if new:
```
techman83: Heh, dictionary vs attribute, still catches me sometimes. I actually wanted to create an object for this, but the Model class seem to require a dict.

HebaruSan:
Welp, that didn't work at all.

techman83:
Well it still needs deploying :wink: (I really need to add those keys to this project) - though the status dumper is a little upset because it doesn't understand the map object.

techman83:
Oh, I see there is a commit to fix the status export already! We probably should have PRs even for small fixes.

Issue #71, opened 2019-10-29
[Bug] Exception on comparing 'last_error'


techman83:
Problem
-------
If there is no 'last_error' in ModStatus an exception is being thrown, we should be able to use a `.get`.

```python
Traceback (most recent call last):
  File ".local/bin/netkan", line 8, in <module>
    sys.exit(netkan())
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli.py", line 79, in indexer
    handler.process_ckans()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/indexer.py", line 242, in process_ckans
    self.process_queue(self.master)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/indexer.py", line 223, in process_queue
    ckan.process_ckan()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/indexer.py", line 155, in process_ckan
    self._process_ckan()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/indexer.py", line 140, in _process_ckan
    if not self.Success and status.last_error != attrs.last_error:
AttributeError: 'dict' object has no attribute 'last_error'
```

HebaruSan:
Someday I will figure out how people manage to be productive in this language.

HebaruSan:
Maybe we could settle for making some of our objects less "Pythonic" by ensuring that their mission critical properties are always present?

techman83:
Well we could do something with the `__getattr__` and we already have our own status class. Something to think about, but a little beyond a bugfix.

HebaruSan:
I'm thinking more like unconditional assignments in `__init__`.

techman83:
[...]

PR #72, opened 2019-10-29
Fix Error Compare

techman83:
This fixes the issue noted in #71. There might be some scope for a more elegant solution on `ModStatus`, however that would require a deeper look at the specifics of PynamoDB.


techman83:
I'll merge this on test pass, that'll get the bot back in action.

PR #76, opened 2019-10-31
Fix development environment

DasSkelett:
This PR includes the changes that where necessary to get my local testing environment running (plus two small tests to get into unit testing for this project):

---

I copied the `dev` service of the `docker-compose.vscode.yml` into the normal `docker-compose.yml`. As far as I understand, it is intended to be used as a "launcher" container, in which you open a console to test e.g. the click commands, or do whatever you want.

---

I removed `pytest` from the development dependencies, we are using `unittests` which is in the stdlib.

---

`init_ssh()` no longer exits the program if no key is provided, so you can use HTTPS cloning if you want.

---

Another test for the download counter was added. Make sure it returns 0 if the kref target is unknown / we didn't implement a way to get the counts from it.
Also I added a test for `cli.clean_cache()`. It's not much, testing `indexer() or `scheduler()` would be better, but those are way harder to write tests for and it's better than nothing.
This should at least prevent things like missing import statements...

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -142,17 +142,3 @@ services:
     volumes:
       - ./netkan:/home/netkan/netkan
     command: download-counter
-  dev:
-    build:
-      context: netkan/.
-      target: dev
-    environment:
-      SSH_KEY: ${CKAN_NETKAN_SSHKEY}
-      NETKAN_REMOTE: ${NETKAN_METADATA_PATH}
-      CKANMETA_REMOTE: ${CKAN_METADATA_PATH}
-      GH_Token: ${CKAN_GH_Token}
-      DISCORD_WEBHOOK_ID: ${DISCORD_WEBHOOK_ID}
-      DISCORD_WEBHOOK_TOKEN: ${DISCORD_WEBHOOK_TOKEN}
-    volumes:
-      - ./netkan:/home/netkan/netkan
-    entrypoint: sleep infinity
--- a/netkan/netkan/utils.py
+++ b/netkan/netkan/utils.py
@@ -17,8 +17,8 @@ def init_repo(metadata, path):
 
 def init_ssh(key, path):
     if not key:
-        logging.error('Private Key required for SSH Git')
-        sys.exit(1)
+        logging.warning('Private Key required for SSH Git')
+        return
     logging.info('Private Key found, writing to disk')
     key_path = Path(path)
     key_path.mkdir(exist_ok=True)
--- a/netkan/setup.py
+++ b/netkan/setup.py
@@ -30,7 +30,6 @@ setup(
     },
     extras_require={
         'development': [
-            'pytest',
             'ptvsd',
             'pylint',
             'autopep8',
--- a/netkan/tests/download_counter.py
+++ b/netkan/tests/download_counter.py
@@ -61,3 +61,12 @@ class TestNetKANNetkanCounts(TestNetKANCounter):
             self.netkan.remote_netkan,
             'http://ksp-ckan.space/netkan/DogeCoinFlag.netkan'
         )
+
+
+class TestNetKANUnknownCounts(TestNetKANCounter):
+
+    def setUp(self):
+        self.netkan = NetkanDownloads(Path(self.test_data, 'NetKAN/UnknownCoinFlag.netkan'), 'token')
+
+    def test_github_repo_api(self):
+        self.assertEqual(self.netkan.get_count(), 0)
--- a/README.md
+++ b/README.md
@@ -1,11 +1,118 @@
-# NetKAN-Infra
+# NetKAN Infrastructure
+
 The NetKAN Infrastructure Project is a re-write of the original monolithic NetKAN-bot, with the goal of easier maintenance, a faster code to deployment pipeline, infrastructure as code.
 
+## Theory of operation
 
-Performing Updates
-------------------
+We use a modern microservice architecture to continuously index new mods and mod updates as the upstream mod files change. The microservices communicate via the AWS SQS system.
+
+### Microservices
+
+#### Inflator
+
+The core of the system, it has the job to _inflate_ NetKAN metadata to CKAN metadata as described in the [Schema] and [Spec].
+
+[Spec]: https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md
+[Schema]: https://github.com/KSP-CKAN/CKAN/blob/master/CKAN.schema
+
+#### Indexer
+
+The Indexer takes inflated metadata and, if the metadata is different to currently existing one, or entirely new, pushes it to the [metadata repository]. It also updates the status database.
+
+[metadata repository]: https://github.com/KSP-CKAN/CKAN-meta
+
+#### Scheduler
+
+Every once in a while, the Scheduler kicks off and submits NetKAN metadata for the Inflator to work with.
+
+#### Webhooks
+
+The webhooks provide a web server that responds to messages from other hosts and services, usually by queueing a module to be inflated by the Inflator.
+
+#### Download Counter
+
+When started, it goes through the entire NetKAN metadata repository and collects the download counts for each of those mods, then commits them to the [metadata repository].
+
+#### Status Page
+
+The status page shows the current status of the inflation for each mod: Time of last inflation, time of last indexing, or errors that occurred during inflation. It accesses a DynamoDB database in the background.
+
+### Containers
+
+Container        | Image                  | Repo            | Code
+---------------- | ---------------------- | --------------- | ----
+Inflator         | kspckan/inflator       | CKAN            | [Netkan/Processors/QueueHandler.cs]
+Scheduler        | kspckan/netkan         | NetKAN-Infra    | [netkan/netkan/scheduler.py]
+Indexer          | kspckan/netkan         | NetKAN-Infra    | [netkan/netkan/indexer.py]
+Download Counter | kspckan/netkan         | NetKAN-Infra    | [netkan/netkan/download_counter.py]
+Status Dumper    | kspckan/netkan         | NetKAN-Infra    | [netkan/netkan/cli.py]
+[Webhooks]       | kspckan/netkan         | NetKAN-Infra    | [netkan/netkan/webhooks]
+Webhooks Proxy   | kspckan/webhooks-proxy | NetKAN-Infra    |
+Cert Bot         | certbot/dns-route53    | certbot/certbot |
+
+[Webhooks]: #Webhooks
+
+[Netkan/Processors/QueueHandler.cs]: https://github.com/KSP-CKAN/CKAN/blob/master/Netkan/Processors/QueueHandler.cs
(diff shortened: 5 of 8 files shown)


docker-compose.yml
```diff
@@ -140,16 +140,29 @@ services:
...
+      INFLATION_SQS_QUEUE: InboundDev.fifo
+      MIRROR_SQS_QUEUE: MirroringDev.fifo
+      STATUS_BUCKET: ckan-test-status
+      STATUS_INTERVAL: 0
       DISCORD_WEBHOOK_ID: ${DISCORD_WEBHOOK_ID}
       DISCORD_WEBHOOK_TOKEN: ${DISCORD_WEBHOOK_TOKEN}
     volumes:
-      - ./netkan:/home/netkan/netkan
+      - .:/home/netkan/workspace
     entrypoint: sleep infinity
```
techman83: `docker-compose.yml` isn't used by VS Code, it's for spinning up a full integration test. It admittedly needs more work and documentation. We should possibly move the `docker-compose.vscode.yml` under `.vscode` and reference it from there.

netkan/netkan/cli.py
```diff
@@ -248,7 +249,8 @@ def download_counter(netkan_remote, ckan_meta, token, key, debug):
         format='[%(asctime)s] [%(levelname)-8s] %(message)s', level=level
     )
     logging.info('Download Counter started at log level %s', level)
-    init_ssh(key, '/home/netkan/.ssh')
+    if netkan_remote.startswith('git@github.com') or ckan_meta.startswith('git@github.com'):
+        init_ssh(key, '/home/netkan/.ssh')
     init_repo(netkan_remote, '/tmp/NetKAN')
```
techman83: I'm not sure adding a check here is great, especially without a comment as to why. I think populating a dummy string in the `docker-compose.yml` is a safer alternative. If using HTTP(s), it'll never try the key.

netkan/Dockerfile
```diff
@@ -8,12 +8,13 @@ RUN python -m unittest -v
 WORKDIR /
 USER root
 RUN rm -Rf /netkan
+ENTRYPOINT [".local/bin/netkan"]
+CMD ["--help"]
```
techman83: Yep, this definitely belongs here. This will pave the way for some build improvements as this layout works, but isn't entirely optimal.

netkan/Dockerfile
```diff
@@ -8,12 +8,13 @@ RUN python -m unittest -v
...
 
 FROM python:3.7 as dev
 RUN useradd -ms /bin/bash netkan
 ADD . /netkan
 WORKDIR /home/netkan
-ADD run_dev.sh /usr/local/bin/
 ENV PATH "$PATH:/home/netkan/.local/bin"
```
techman83: We should restore the `run_dev.sh` and the dev `ENTRYPOINT` that was dropped during the VS Code stuff.

netkan/tests/download_counter.py
```diff
@@ -61,3 +61,12 @@ def test_remote_netkan(self):
...
         )
+
+
+class TestNetKANUnknownCounts(TestNetKANCounter):
+
+    def setUp(self):
+        self.netkan = NetkanDownloads(Path(self.test_data, 'NetKAN/UnknownCoinFlag.netkan'), 'token')
+
+    def test_github_repo_api(self):
+        self.assertEqual(self.netkan.get_count(), 0)
```
techman83: I :heart: tests!!!

techman83:
One thing I have been doing on some of my work projects is being a bit smarter with the multi-stage stuff. Here's a patch you can look at for doing the same here

```patch
diff --git a/netkan/Dockerfile b/netkan/Dockerfile
index 565a29f..c80d4b2 100644
--- a/netkan/Dockerfile
+++ b/netkan/Dockerfile
@@ -3,27 +3,30 @@ RUN useradd -ms /bin/bash netkan
 ADD . /netkan
 WORKDIR /netkan
 USER netkan
-RUN pip install --user .
+RUN pip install --user . --no-warn-script-location
 RUN python -m unittest -v
 
-FROM python:3.7 as dev
-RUN useradd -ms /bin/bash netkan
-ADD . /netkan
-WORKDIR /home/netkan
-ADD run_dev.sh /usr/local/bin/
-ENV PATH "$PATH:/home/netkan/.local/bin"
-USER netkan
-RUN pip install --user /netkan/.[development]
-
-FROM python:3.7
+FROM python:3.7 as production
 COPY --from=base /home/netkan /home/netkan
 RUN useradd -Ms /bin/bash netkan
 RUN chown -R netkan:netkan /home/netkan
 WORKDIR /home/netkan
 USER netkan
 ADD .gitconfig .
 ENTRYPOINT [".local/bin/netkan"]
 CMD ["--help"]
+
+FROM production as dev
+USER root
+ADD . /netkan
+WORKDIR /home/netkan
+USER netkan
+RUN pip install --user /netkan/.[development]
+ENTRYPOINT ["/usr/local/bin/run_dev.sh"]
+
+from production
```
Assuming that the docker engine in travis is new than 18.03, we can pass `--target production` and it will skip building the development stage entirely. Which will speed the build up slightly.

netkan/netkan/cli.py
DasSkelett: What if I just remove the `sys.exit(1)` in `init_ssh()`?
A missing ssh key is not a fatal error, especially if a HTTPS URL is used.

docker-compose.yml
DasSkelett: But it does basically the same, doesn't it? It starts up the container and goes to idle, waiting for a user to open a remote console for testing stuff?

DasSkelett:
Rebase the PR on top of #78:
* Removed the checks before `init_ssh()`, but `return` instead of `sys.exit(1)` if no SSH key is submitted.
* Removed the changes from `netkan/Dockerfile` since they are now obsolete
* I accidentally pushed my basic documentation attempt too as I just realized. If you want to keep it, I let the commit stay, else I'd remove it from this PR.

docker-compose.yml
techman83: Yeah, I added it before I implemented all the bits and pieces for VS Code. This stub here will never get used by VS code. But they share a project namespace for convenience.

netkan/netkan/utils.py
```diff
@@ -17,8 +17,8 @@ def init_repo(metadata, path):
 
 def init_ssh(key, path):
     if not key:
-        logging.error('Private Key required for SSH Git')
-        sys.exit(1)
+        logging.warning('Private Key required for SSH Git')
+        return
```
techman83: Yeah, this seems logical to me.

netkan/setup.py
```diff
@@ -10,6 +10,7 @@
     author_email='techman83@gmail.com',
     packages=find_packages(),
     install_requires=[
+        'wheel',
         'boto3',
```
techman83: Why are we adding 'wheel' as an install dependency here?

netkan/setup.py
```diff
@@ -30,7 +31,6 @@
     },
     extras_require={
         'development': [
-            'pytest',
             'ptvsd',
```
techman83: Have you checked if VS Code uses this? (I don't think it does, I'll have to check).

netkan/setup.py
DasSkelett: Oh right. Pip tries to build wheels when installing all the dependencies.
`wheel` is already preinstalled in the Python3 Docker Container, but not in a venv for example, and pip will throw errors (or warnings, don't know anymore) if `wheel` isn't installed.
This affects the containers not at all (pip says dependency already satisfied), but helps venv users (I have a venv setup too, because my IDE likes having the imported packages installed for code inspection).

netkan/setup.py
DasSkelett: `.vscode/settings.json` contains the following lines:
```json
  "python.testing.pytestEnabled": false,
  "python.testing.nosetestsEnabled": false,
  "python.testing.unittestEnabled": true,
```

So VSCode shouldn't use pytest. Additionally, all our tests are set up for `unittest`, I don't know if `pytest` could even run them.

However I can't test it, I don't use VS Code.

docker-compose.yml
DasSkelett: Is it used by anything else? Should I delete it entirely?

techman83: I don't think we need to stress too hard on this PR. We probably should have a rethink of the dev workflow, since you'll be more likely to use this I'm happy for you to take the lead.

netkan/setup.py
techman83: I still use a venv for the project. Though it makes no difference to the container, adding a build dependency as something the project requires seems a bit wrong.

> install_requires is a setuptools setup.py keyword that should be used to specify what a project minimally needs to run correctly. When the project is installed by pip, this is the specification that is used to install its dependencies.

Maybe in the dev dependencies is more appropriate?

netkan/setup.py
techman83: Nope, not required by VS Code by the looks. PyTest _can_ run unittest TestCase stuff, but we can worry about it if we want to use it later.

> pytest supports running Python unittest-based tests out of the box. It’s meant for leveraging existing unittest-based test suites to use pytest as a test runner and also allow to incrementally adapt the test suite to take full advantage of pytest’s features.

docker-compose.yml
DasSkelett: So just to make sure, this is not used by you (or @HebaruSan?) or any testing stuff?

Because if not, I'd remove it, I don't need it either. But I don't want to break anyones workflow there.

netkan/setup.py
DasSkelett: I remember I added it because I thought it pip threw errors due to this and didn't want to install the other packages.
But checking back I'm not sure why it threw errors, it doesn't throw them anymore when I delete the venv folder and install again :man_shrugging:

I'm going to remove it again, or would you want to keep it in the dev deps?

docker-compose.yml
techman83: I have no strong opinions either way. The infrastructure is bedded down enough where this probably isn't needed and anything else we do should really be adding tests.

netkan/setup.py
techman83: Yeah, I'd just drop it. It's neither a package dependency or a development one.

docker-compose.yml
HebaruSan: If I'm using this somehow, I don't know about it. I'm very much following the lead of whatever you folks put in the files when it comes to docker.

DasSkelett:
Okay, so I removed the `dev` service from docker-compose and the `wheel` dependency from `setup.py`.

The `clean_cache` will fail locally for you (and me), because it tests on code that contains hardcoded pathes (`/home/netkan/ckan_cache`).
I plan to replace those withenv vars or something, if you have an idea, bring it on. But I would do that in another PR.

This PR is good-to-merge for me.

techman83:
Awesome work, thank you so much for doing this! I don't see any reason not to merge this now.

PR #77, opened 2019-11-04
Put long Discord messages in code blocks

HebaruSan:
## Problem

As of #74 we are sending uncaught exceptions to Discord. Stack traces look a little weird:

image

## Cause

Discord uses markdown formatting, in which `*` and `**` begin and end italics and bold respectively. The stack trace messages contain these characters but are not attempting to use them for formatting.

## Changes

Now if a message contains a newline, we add a `` ``` `` triple backtick prefix and suffix to force it to display as a code block. This should prevent special characters from being treated as formatting.

image

One-line messages will be left as-is to avoid spamming the chat with big dark boxes.


netkan/netkan/notifications.py
```diff
@@ -19,7 +19,8 @@ def __init__(self, webhook_id, webhook_token):
                                                adapter=discord.RequestsWebhookAdapter())
 
     def emit(self, record):
-        self.webhook.send(self.format(record))
+        fmt = self.format(record)
+        self.webhook.send(f'```{fmt}```' if "\n" in fmt else fmt)
```
techman83: It also accepts markdown styling, so we could do `f'```python\n{fmt}```'`

HebaruSan: True, but the strings won't be Python code. Even a Python stack trace isn't valid code such that a syntax highlighter would treat it correctly.

techman83: It does improve readability of the message even if it isn't strictly Python.

techman83:
I'll add the automated deployment stuff to this PR later today and merge it. I'm all for making things more readable!

techman83:
Well that would have caused a surprise if I hadn't noticed that on something else I was working on.

https://github.com/boto/botocore/commit/e87e7a745fd972815b235a9ee685232745aa94f9

After I merge that should build, push + redeploy the services!

PR #78, opened 2019-11-06
Build Improvements

techman83:
This PR cleans up Netkan's docker build, ensuring dev is a downstream consumer of production and only production image gets built and pushed. VS Code's compose file is moved into the `.vscode` directory.

Production's stack also includes IAM to allow the builder to update our status files via Travis.

### DIFF
--- a/netkan/Dockerfile
+++ b/netkan/Dockerfile
@@ -21,7 +21,6 @@ FROM production as dev
 USER root
 ADD . /netkan
 ADD run_dev.sh /usr/local/bin/
-WORKDIR /home/netkan
 USER netkan
 RUN pip install --user /netkan/.[development]
 ENTRYPOINT ["/usr/local/bin/run_dev.sh"]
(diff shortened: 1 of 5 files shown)


techman83:
That's ok, even just second set of eyes will pickup stuff like typos. I'll merge it in later so I can make sure the deployment is good (it's quite a change to the build).

DasSkelett:
Thanks @techman83, this should give me a good starting point to rebase my PR.

netkan/Dockerfile
```diff
@@ -3,27 +3,27 @@ RUN useradd -ms /bin/bash netkan
...
 ADD .gitconfig .
+ENV PATH "$PATH:/home/netkan/.local/bin"
 ENTRYPOINT [".local/bin/netkan"]
 CMD ["--help"]
+
+FROM production as dev
+USER root
+ADD . /netkan
+ADD run_dev.sh /usr/local/bin/
+WORKDIR /home/netkan
```
DasSkelett: The `WORKDIR` is already set to `/home/netkan` in `production`, do we need it twice?

techman83: Indeed we don't, good spot :slightly_smiling_face:

PR #79, opened 2019-11-06
Schedule inflation of webhook-only mods once a day

HebaruSan:
## Motivation

In #40 we suppressed scheduling of modules that are adequately handled by webhook notifications.

In normal operation this helpfully reduces resource consumption, but sometimes we develop things like #70 that add functionality that depends on fresh inflations, which leaves the excluded modules behind. And there's always the possibility that we miss a webhook inflation due to some glitch on our side or SpaceDock's.

## Changes

Now modules that are excluded from normal passes will be scheduled for inflation once per day.

### DIFF
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -107,18 +107,27 @@ def indexer(queue, metadata, token, repo, user, key,
     help='Disable AWS Credit Checks',
 )
 @click.option(
-    '--schedule-all', is_flag=True, default=False,
-    help='Schedule all modules even if we think they should be covered by webhooks'
+    '--group',
+    type=click.Choice(['all', 'webhooks', 'nonhooks']), default="nonhooks",
+    help='Which mods to schedule'
 )
-def scheduler(queue, netkan_remote, max_queued, debug, dev, schedule_all):
+@click.option(
+    '--min-credits', default=200,
+    help='Only schedule if we have at least this many credits remaining'
+)
+def scheduler(queue, netkan_remote, max_queued, debug, dev, group, min_credits):
     level = logging.DEBUG if debug else logging.INFO
     logging.basicConfig(
         format='[%(asctime)s] [%(levelname)-8s] %(message)s', level=level
     )
     logging.info('Scheduler started at log level %s', level)
 
-    sched = NetkanScheduler(Path('/tmp/NetKAN'), queue, force_all=schedule_all)
-    if sched.can_schedule(max_queued, dev):
+    sched = NetkanScheduler(
+        Path('/tmp/NetKAN'), queue,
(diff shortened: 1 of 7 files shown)


techman83:
This would certainly run the scheduler every Monday at  2am. But there are a couple of things that I think we need to address.

- In hindsight there isn't a reason to schedule _all_ the netkans. I did have some code for inverting the logic based on a '--webhooks-only' flag. But it was messy and I haven't had a chance to get back and have a closer look.
- This may not fire at all as the status check is hardcoded to requiring 100 credits. This doesn't matter with the regular scheduling, but we're hoping we'll have spare credits when this schedule fires. A '--credits' flag was something I was also working with above. Default to 200, allow the webhooks to run with 100.
- I think we should run this daily, but that's a trivial implementation detail.

HebaruSan:
- Replaced the `--schedule-all` flag with a `--group` option that can be `webhooks`, `nonhooks`, or `all`
- Added a `--min-credits` option
- Changed to `rate(1 day)`

techman83:
From my brief read over this looks good. I'll take a closer look after work and merge it in/update the stack.

netkan/netkan/cli.py
```diff
@@ -107,18 +107,27 @@ def indexer(queue, metadata, token, repo, user, key,
     help='Disable AWS Credit Checks',
 )
 @click.option(
-    '--schedule-all', is_flag=True, default=False,
-    help='Schedule all modules even if we think they should be covered by webhooks'
+    '--group',
+    type=click.Choice(['all', 'webhooks', 'nonhooks']), default="nonhooks",
+    help='Which mods to schedule'
```
techman83: Click choices are really cool! :+1:

netkan/netkan/scheduler.py
```diff
@@ -25,29 +27,24 @@ def netkans(self):
...
             'QueueUrl': self.queue_url,
             'Entries': batch
         }
 
+    def _in_group(self, netkan):
+        if netkan.hook_only():
+            return self.webhooks_group
+        else:
+            return self.nonhooks_group
+
```
techman83: I'm not overly sure about this logic, but we can add some tests for this and re-factor later if need be. Much better than what was springing to mind when I last had a look and this does the job. :+1:

prod-stack.py
```diff
@@ -330,7 +330,7 @@
 # the tasks.
 scheduler_resources = []
 for task in [
-        'Scheduler', 'CertBot', 'StatusDumper', 'DownloadCounter']:
+        'Scheduler', 'SchedulerFullPass', 'CertBot', 'StatusDumper', 'DownloadCounter']:
     scheduler_resources.append(Sub(
```
techman83: I'll rename this before merging/deploying as it's not really a full pass any more.

PR #82, opened 2019-11-10
Always schedule the webhook pass

HebaruSan:
## Problem

Currently the webhooks pass from #79 won't run if there are 20 or more messages in the inflation queue. This can happen anytime the webhooks pass starts while the main pass is in progress, which I think happened last time.

## Changes

Now we pass `--max-queued 2000` to ensure that the daily pass won't be missed.
(Should we use a special value like `-1` to force it to run regardless? Let me know, I was going for minimal changes but that sort of thing might make sense too.)

In addition, a TODO item that should have been pruned from `scheduler.py` in #79 is trimmed down to only info that's still applicable, and linter fixes requested by VSCode are applied in that file. (If they look familiar, that's because they were initially done in #58 and then rolled back).

### DIFF
--- a/netkan/netkan/scheduler.py
+++ b/netkan/netkan/scheduler.py
@@ -1,8 +1,9 @@
-import boto3
 import datetime
 import logging
-import requests
 from pathlib import Path
+import boto3
+import requests
+
 from .metadata import Netkan
 from .common import sqs_batch_entries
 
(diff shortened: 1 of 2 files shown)


techman83:
Thanks for adjusting the comment and that looks all good. I think 2000 is an appropriate number, if we've got more than that in the queue, something has really gone wrong.

PR #86, opened 2019-11-11
Restore Console Logging

techman83:
Console logging was lost in #77, this restores console logging and makes it consistent across all the commands.

The only drawback being that the `--debug` flag is on the group and must be passed before the command, ie
```bash
netkan@bef8cb5997ae:~/workspace$ netkan --debug indexer --help
[2019-11-11 09:14:41,812] [INFO    ] Logging started for '--debug' at log level 10
Usage: netkan indexer [OPTIONS]

Options:
  --queue TEXT       SQS Queue to poll for metadata
  --metadata TEXT    Path/URL/SSH to Metadata Repo
  --token TEXT       GitHub Token for PRs  [required]
  --repo TEXT        GitHub repo to raise PR against (Org Repo: CKAN-meta)
  --user TEXT        GitHub user/org repo resides under (Org User: KSP-CKAN)
  --timeout INTEGER  Reduce message visibility timeout for testing
  --key TEXT         [required]
  --help             Show this message and exit.
netkan@bef8cb5997ae:~/workspace$ netkan indexer --debug --help
[2019-11-11 09:14:48,690] [INFO    ] Logging started for 'indexer' at log level 20
Usage: netkan indexer [OPTIONS]
Try "netkan indexer --help" for help.

Error: no such option: --debug
```
There are few ways to work around it, but it's only useful in dev and rare to need it even then. So I'm comfortable to leave it as is unless it becomes annoying.

### DIFF
--- a/netkan/netkan/notifications.py
+++ b/netkan/netkan/notifications.py
@@ -23,7 +23,13 @@ class DiscordLogHandler(logging.Handler):
         self.webhook.send(f'```{fmt}```' if "\n" in fmt else fmt)
 
 
-def setup_log_handler():
+def setup_log_handler(debug=False):
+    level = logging.DEBUG if debug else logging.INFO
+    logging.basicConfig(
+        format='[%(asctime)s] [%(levelname)-8s] %(message)s', level=level
+    )
+    logging.info('Logging started for \'%s\' at log level %s', sys.argv[1], level)
+
     # Set up Discord logger so we can see errors
     discord_webhook_id = os.environ.get('DISCORD_WEBHOOK_ID')
     discord_webhook_token = os.environ.get('DISCORD_WEBHOOK_TOKEN')
--- a/netkan/netkan/notifications.py
+++ b/netkan/netkan/notifications.py
@@ -24,11 +24,12 @@ class DiscordLogHandler(logging.Handler):
 
 
 def setup_log_handler(debug=False):
-    level = logging.DEBUG if debug else logging.INFO
-    logging.basicConfig(
-        format='[%(asctime)s] [%(levelname)-8s] %(message)s', level=level
-    )
-    logging.info('Logging started for \'%s\' at log level %s', sys.argv[1], level)
+    if not sys.argv[0].endswith('gunicorn'):
+        level = logging.DEBUG if debug else logging.INFO
+        logging.basicConfig(
+            format='[%(asctime)s] [%(levelname)-8s] %(message)s', level=level
+        )
+        logging.info('Logging started for \'%s\' at log level %s', sys.argv[1], level)
 
     # Set up Discord logger so we can see errors
     discord_webhook_id = os.environ.get('DISCORD_WEBHOOK_ID')
--- a/netkan/Dockerfile
+++ b/netkan/Dockerfile
@@ -5,6 +5,7 @@ WORKDIR /netkan
 USER netkan
 RUN pip install --user . --no-warn-script-location
 RUN python -m unittest -v
+RUN /home/netkan/.local/bin/netkan --help
 
 FROM python:3.7 as production
 COPY --from=base /home/netkan /home/netkan
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -16,10 +16,14 @@ from .download_counter import DownloadCounter
 from .ticket_closer import TicketCloser
 
 
+@click.option(
+    '--debug', is_flag=True, default=False,
+    help='Enable debug logging',
+)
 @click.group()
-def netkan():
+def netkan(debug):
     # Set up Discord logger so we can see errors
-    if setup_log_handler():
+    if setup_log_handler(debug):
         # Catch uncaught exceptions and log them
         sys.excepthook = catch_all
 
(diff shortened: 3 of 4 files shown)


netkan/netkan/notifications.py
```diff
@@ -23,7 +23,13 @@ def emit(self, record):
         self.webhook.send(f'```{fmt}```' if "\n" in fmt else fmt)
 
 
-def setup_log_handler():
+def setup_log_handler(debug=False):
+    level = logging.DEBUG if debug else logging.INFO
+    logging.basicConfig(
+        format='[%(asctime)s] [%(levelname)-8s] %(message)s', level=level
+    )
+    logging.info('Logging started for \'%s\' at log level %s', sys.argv[1], level)
```
HebaruSan: `sys.argv[1]` probably isn't defined for the webhooks.

techman83: It is, it's just not useful :slightly_smiling_face: 
```bash
netkan@bef8cb5997ae:~/workspace$ gunicorn -b 0.0.0.0:5000 --access-logfile - "netkan.webhooks:create_app()"
[2019-11-12 01:44:06 +0000] [2344] [INFO] Starting gunicorn 19.9.0
[2019-11-12 01:44:06 +0000] [2344] [INFO] Listening at: http://0.0.0.0:5000 (2344)
[2019-11-12 01:44:06 +0000] [2344] [INFO] Using worker: sync
[2019-11-12 01:44:06 +0000] [2347] [INFO] Booting worker with pid: 2347
[2019-11-12 01:44:09,108] [INFO    ] Logging started for '-b' at log level 20
```

netkan/netkan/notifications.py
```diff
@@ -23,7 +23,14 @@ def emit(self, record):
         self.webhook.send(f'```{fmt}```' if "\n" in fmt else fmt)
 
 
-def setup_log_handler():
+def setup_log_handler(debug=False):
+    if not sys.argv[0].endswith('gunicorn'):
```
HebaruSan: Gunicorn is one of several options for WSGI servers, our code should not care which one we are using.

netkan/netkan/notifications.py
```diff
@@ -23,7 +23,14 @@ def emit(self, record):
...
+        level = logging.DEBUG if debug else logging.INFO
+        logging.basicConfig(
+            format='[%(asctime)s] [%(levelname)-8s] %(message)s', level=level
+        )
+        logging.info('Logging started for \'%s\' at log level %s', sys.argv[1], level)
```
HebaruSan: This is giving wrong output already in your sample command, this should say `Logging started for 'indexer'`:

```
[2019-11-11 09:14:41,812] [INFO    ] Logging started for '--debug' at log level 10
```

HebaruSan: Maybe we don't need to repeat back the command name to the user, and this could just state the logging level?

HebaruSan:
How exactly was console logging lost? I'd like to check over that part of the fix.

netkan/netkan/notifications.py
techman83: Considering we own and deploy the code on our infrastructure. As it's just for making console logging a little nicer, I don't think it matters too much.

netkan/netkan/notifications.py
techman83: When you're running on ECS, it's pretty easy to pick the wrong container.
```
$ docker logs -f f341ec0a3918
# github.com:22 SSH-2.0-babeld-2e9d163d
```
It's something that will stand out if you picked the wrong one.

techman83:
> How exactly was console logging lost? I'd like to check over that part of the fix.

Because we're initialising the logger to do the discord notifications. The way the logger class works, the first initialization wins. Means if you consume something that has its own logging enabled, it picks up the existing settings rather than blatting them.

The only things that get printed now are stuff destined for stdout/stderr (as seen in my example about docker).

techman83:
Missed that the setup for the new dependency was incorrect, so I've added and will merge once the build succeeds.
```python
  File ".local/bin/netkan", line 5, in <module>
    from netkan.cli import netkan
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli.py", line 16, in <module>
    from .ticket_closer import TicketCloser
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/ticket_closer.py", line 3, in <module>
    import github
ModuleNotFoundError: No module named 'github'
```
Also a simple sanity check in the first stage that will fail on things like this.
```bash
Step 8/25 : RUN /home/netkan/.local/bin/netkan --help
 ---> Running in 084a826fbd1e
  File "/home/netkan/.local/bin/netkan", line 5, in <module>
    from netkan.cli import netkan
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli.py", line 16, in <module>
    from .ticket_closer import TicketCloser
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/ticket_closer.py", line 3, in <module>
    import github
ModuleNotFoundError: No module named 'github'
ERROR: Service 'indexer' failed to build: The command '/bin/sh -c /home/netkan/.local/bin/netkan --help' returned a non-zero code: 1
```

PR #91, opened 2019-11-14
Push master before staged branches

### LINKED ISSUE #83: [Bug] Staging process including extra commits
Noted in KSP-CKAN/CKAN-meta#1530

HebaruSan:
## Problem

KSP-CKAN/CKAN-meta#1530 staged one mod, but displayed nine in its list of commits:

image

This is confusing for pull request reviewers because it looks like more mods have been staged than should have been (although in fact no such problem has occurred, see link below).

## Cause

See https://github.com/KSP-CKAN/NetKAN-Infra/issues/83#issuecomment-553681260; to sum up, we are giving GitHub a "preview" of future commits to master when we push a staging branch.

## Changes

Now we push master after we finish processing master messages and before we process staged messages. The `update_master(push=True)` call is moved from `__exit__` to `process_ckans`.

Note that previously the call to `update_master` in `__exit__` provided a strong guarantee that master would always be pushed when a `with MessageHandler` block ended regardless of which calls were performed within it. For example, another class could have instantiated a `MessageHandler` and then called `process_queue` itself to handle a bunch of messages in some other ordering, relying on `__exit__` to push them. With the push moved to `process_ckans`, this would no longer work. To guard against this, some functions of `MessageHandler` are renamed with an underscore at the beginning to express the notion that they should only be called by `MessageHandler` itself. This means that `append` and `process_ckans` are now the only things that other classes should be calling, which will work as intended.

Fixes #83.

### DIFF
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -204,11 +204,10 @@ class MessageHandler:
     # we can ensure we call close on it and run our handler inside
     # a context manager
     def __enter__(self):
-        self.update_master()
+        self._update_master()
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self.update_master(push=True)
         self.repo.close()
 
     def append(self, message):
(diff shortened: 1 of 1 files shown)


techman83:
I think that's all reasonable. If we fail to exit properly an exception occurred somewhere. In which case the whole process terminates and a new container will be spawned with a fresh clone of the repo.

PR #92, opened 2019-11-14
Break up cli.py

HebaruSan:
## Motivation

`cli.py` is getting pretty big, which makes for less convenient maintenance: scrolling back and forth to find relevant code, potential conflicting imports, resolving conflicts, etc.

## Changes

Now each subcommand from `cli.py` is split into its own new, much shorter file in a `netkan.cli` module.
`cli/__init__.py` imports all of them and builds the `netkan` group.

This makes each subcommand's code more self-contained and isolated.


techman83:
Hmm, we do need to do some refactoring to make our cli entrypoint a little cleaner, but rather than breaking it up we could simplify it. Ideally it should be a bunch of stubs, but it was always my intention to revisit it rather than get bogged down while we replaced the old infrastructure.

I have been pondering some ideas on how to do this to reduce our clutter and keep a single file. I might take a look after work tomorrow night.
- Write a custom click decorator that contains all our common arguments and picks out the ones we want based on a list, ie `@netkan.options(cli_options=['outbound_queue', 'inbound_queue']), which would end up as named args on the consuming cli.
- Make an inheritable SQS runner class, so we can make a class method available to all our classes that listen on queues (something I think that's needed for the Adder + Mirroring)

HebaruSan:
Well, what's the downside to putting each subcommand in its own file?

techman83:
It adds a lot of boiler plate and a lot of places we would need to change stuff if we decide to refactor a component. Operationally it makes no difference.

HebaruSan:
> It adds a lot of boiler plate and a lot of places we would need to change stuff if we decide to refactor a component.

Huh, I'm not seeing that at all. The net change in line count is +36: blank lines, `import` statements that are no longer shared between unrelated pieces of code, and importing the subcommands into `__init__.py`. The `@click.option`-type boiler plate is unchanged.
image

And refactoring a component would be no harder, possibly easier, since we don't have to worry about accidentally affecting an unrelated subcommand that's sharing the file that we're editing. Just open the file for the subcommand we care about and go.

techman83:
It might be better if I show you with code. I'm not necessarily opposed to breaking it apart like this, but I don't think that's where we should start.

HebaruSan:
#122 is splitting the commands into cli/services.py and cli/utilities.py, probably a good direction to take it.

PR #94, opened 2019-11-14
Make Scheduler sorting case insensitive

HebaruSan:
## Problem

After #93 we are sorting the modules for scheduling, however the sort is case sensitive so after `ZZZRadioTelescope` we process a bunch of modules that start with a lowercase letter:

image

## Changes

Now the sort is case insensitive, so lowercase letters will be folded in with their uppercase counterparts.

Python provides `str.casefold` for this, which is like `str.lower`, but it converts `ß` to `ss`. It doesn't accept a `Path` parameter, so we use a `lambda` to casefold the stem.

### DIFF
--- a/netkan/netkan/scheduler.py
+++ b/netkan/netkan/scheduler.py
@@ -26,7 +26,8 @@ class NetkanScheduler:
     def netkans(self):
         # This can easily be recursive with '**/*.netkan', however
         # implementing like for like initially.
-        return (Netkan(f) for f in sorted(self.path.glob('*.netkan')))
+        return (Netkan(f) for f in sorted(self.path.glob('*.netkan'),
+                                          key=lambda p: p.stem.casefold()))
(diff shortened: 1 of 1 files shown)


netkan/netkan/scheduler.py
```diff
@@ -26,7 +26,8 @@ def __init__(self, path, queue, base='NetKAN/', nonhooks_group=False, webhooks_g
     def netkans(self):
         # This can easily be recursive with '**/*.netkan', however
         # implementing like for like initially.
-        return (Netkan(f) for f in sorted(self.path.glob('*.netkan')))
+        return (Netkan(f) for f in sorted(self.path.glob('*.netkan'),
+                                          key=lambda p: p.stem.casefold()))
```
techman83: Cool! Today I learned about case fold :raised_hands:

PR #96, opened 2019-11-16
Automatically freeze modules after 3 years without updates

HebaruSan:
## Motivation

Sorting http://status.ksp-ckan.space/ by Last Indexed, it's apparent that hundreds of indexed mods haven't been updated in 3 years or longer (maybe 40% of the total going by the scrollbar). These probably will never be updated (either because they don't need updates or because they're abandoned), yet we crawl them every day, some several times per day. This incurs extra CPU, network, and possibly RAM consumption.

## Changes

Now once a week we run a new `auto-freezer` command that checks whether any of the currently indexed mods have a `ModStatus.last_indexed` value 1000 days ago or earlier (which is pretty close to 3 years). (~~`None` is treated as infinitely old, since it means the mod has not been indexed since we started tracking its status.~~ This *would* be true, except that the legacy Perl webhooks never updated the status, so a `None` value is unreliable. See #97 for help with that.) All such mods are renamed to `identifier.frozen` on a `freeze/auto` branch, one commit per mod. This branch is pushed to GitHub and a pull request is created to allow team members to manually review the results and merge to master if appropriate. This will allow us to reduce the workload of the bot over time.

To support pushing to KSP-CKAN/NetKAN, its remote string is updated to SSH format.

In addition, a new `ModStatus.frozen` property exists. After its normal pass, the auto freezer command will set this property to `True` for mods that don't have an active .netkan file. This will allow us to create a filter on the status page to hide such mods.

## Help wanted

Things I couldn't figure out:

- [ ] The `freeze/auto` branch needs to have new mods added to it as they're added to master, ideally without losing previous work, something like `git rebase master`. But I couldn't figure out how to do that in GitPython.
- [ ] If we decide not to freeze a particular idle mod, the script may keep re-submitting PRs to freeze it. Ideally we would create a way to suppress this.

Any help would be very much appreciated.

### DIFF
--- a/netkan/netkan/auto_freezer.py
+++ b/netkan/netkan/auto_freezer.py
@@ -1,0 +1,74 @@
+import logging
+import datetime
+from pathlib import Path
+import git
+
+from .status import ModStatus
+
+
+class AutoFreezer:
+
+    BRANCH_NAME = 'freeze/auto'
+    UNFROZEN_SUFFIX = 'netkan'
+    FROZEN_SUFFIX = 'frozen'
+
+    def __init__(self, netkan_repo, github_pr):
+        self.netkan_repo = netkan_repo
+        self.github_pr = github_pr
+
+    def freeze_idle_mods(self, days_limit):
+        update_cutoff = datetime.datetime.now() - datetime.timedelta(days=days_limit)
+        self.netkan_repo.remotes.origin.pull('master', strategy_option='ours')
+        ids_to_freeze = [ident for ident in self._ids() if self._too_old(ident, update_cutoff)]
+        if ids_to_freeze:
+            self._checkout_branch(self.BRANCH_NAME)
+            for ident in ids_to_freeze:
+                if Path(self.netkan_repo.working_dir,
+                        'NetKAN',
+                        f'{ident}.{self.UNFROZEN_SUFFIX}'
+                        ).exists():
+                    logging.info('Freezing %s', ident)
+                    self._add_freezee(ident)
+                else:
+                    logging.info('Already froze %s', ident)
+            self._submit_pr(self.BRANCH_NAME, days_limit)
+            self.netkan_repo.heads.master.checkout()
+
+    def _checkout_branch(self, name):
+        try:
+            self.netkan_repo.remotes.origin.fetch(name)
+        except git.GitCommandError:
+            logging.info('Unable to fetch %s', name)
+
+        (getattr(self.netkan_repo.heads, name, None)
+         or self.netkan_repo.create_head(name)).checkout()
+
+    def _ids(self):
+        return (nk_path.stem for nk_path
+                in Path(self.netkan_repo.working_dir, "NetKAN").glob('*.netkan'))
+
+    def _too_old(self, ident, update_cutoff):
+        status = ModStatus.get(ident)
+        last_indexed = getattr(status, 'last_indexed', None)
+        if not last_indexed:
+            # Never indexed since the start of status tracking = 4+ years old
+            return True
+        else:
+            return last_indexed < update_cutoff
+
+    def _add_freezee(self, ident):
+        self.netkan_repo.index.move([
+            Path('NetKAN', f'{ident}.{self.UNFROZEN_SUFFIX}').as_posix(),
+            Path('NetKAN', f'{ident}.{self.FROZEN_SUFFIX}').as_posix()
+        ])
+        self.netkan_repo.index.commit(f'Freeze {ident}')
+
+    def _submit_pr(self, branch_name, days):
+        logging.info('Submitting pull request for %s', branch_name)
+        self.netkan_repo.remotes.origin.push(f'{branch_name}:{branch_name}')
+        self.github_pr.create_pull_request(
+            branch=branch_name,
+            title='Freeze idle mods',
+            body=(f'The attached mods have not updated in {days} or more days.'
+                  ' Freeze them to save the bot some CPU cycles.'),
+        )
--- a/netkan/netkan/auto_freezer.py
+++ b/netkan/netkan/auto_freezer.py
@@ -52,7 +52,8 @@ class AutoFreezer:
         last_indexed = getattr(status, 'last_indexed', None)
         if not last_indexed:
             # Never indexed since the start of status tracking = 4+ years old
-            return True
+            # ... except for mods that were updated by the old webhooks :(
+            return False
         else:
             return last_indexed < update_cutoff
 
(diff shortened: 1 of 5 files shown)


DasSkelett:
What if it creates a new branch for each mod to freeze? This would solve the issue of the rebase.
And it could be used to solve the second problem too, though not very pretty:
Don't delete the freeze branches in the upstream repo (KSP_CKAN/NetKAN) if the PR is "denied", and let the bot use this as indication to not create another PR.

However this could clutter the repository with unused branches over time :/

HebaruSan:
> What if it creates a new branch for each mod to freeze?

I don't relish the notion of spamming the repo with hundreds of pull requests on the first pass.

techman83:
> > What if it creates a new branch for each mod to freeze?
> 
> I don't relish the notion of spamming the repo with hundreds of pull requests on the first pass.

I'll review this properly when I wake up properly, but you do know my opinion already on making sacrifices to operational code relating to historical data :wink: 

We probably should list them, do 90% in single hand crafted PR and let the new task munch on the rest.

DasSkelett:
```sh
cat netkan.json | jq 'to_entries | .[] | select(.value.last_indexed == null) | .key' | tr -d '"' > never.list
cat netkan.json | jq 'to_entries | .[] | select(.value.last_indexed and (.value.last_indexed | startswith("2015"))) | .key ' | tr -d '"' > 2015.list
cat netkan.json | jq 'to_entries | .[] | select(.value.last_indexed and (.value.last_indexed | startswith("2016"))) | .key ' | tr -d '"' > 2016.list
cat netkan.json | jq 'to_entries | .[] | select(.value.last_indexed and (.value.last_indexed | startswith("2017"))) | .key ' | tr -d '"' > 2017.list
cat netkan.json | jq 'to_entries | .[] | select(.value.last_indexed and (.value.last_indexed | startswith("2018"))) | .key ' | tr -d '"' > 2018.list
cat netkan.json | jq 'to_entries | .[] | select(.value.last_indexed and (.value.last_indexed | startswith("2019"))) | .key ' | tr -d '"' > 2019.list

wc -l *
> 315  never.list
> 10   2015.list
> 243  2016.list
> 96   2017.list
> 132  2018.list
> 1000 2019.list
```
(Thanks @HebaruSan for letting me know that tool, it's great!)

There are indeed quite a few mods to start with.
A PR beforehand makes sense. But we have to leave some for the auto-freezer ;)


Also could we use the auto-freeer to prune all the frozen netkans (including manually frozen ones) from the status db?

HebaruSan:
Nice! I forgot we could get the `ModStatus` data from the status page's netkan.json. This lists the modules with a date before a given date (apparently there are 253 before 2017-01-01):

`jq '("2017-01-01T00:00:00Z"|fromdate) as $cutoff | with_entries(select(.value.last_indexed and (.value.last_indexed[0:19]+"Z"|fromdate)<$cutoff) | {key, value:.value.last_indexed})' < netkan.json`

Also it would be nice to recover the `last_indexed` values for modules that were only indexed by the old web hooks. We could extract it from CKAN-meta's git log and then set it in `ModStatus`.

HebaruSan:
Added a one-off script to backfill `last_indexed` dates. This should probably run first, since having the timestamps filled in will make the auto-freezer more effective.

HebaruSan:
`ModStatus.frozen` property added in latest commit (rebased to get the timestamp import script so I could copy pieces of it).

techman83:
In hindsight, a single branch is probably ok. As long as our workflow is to delete the branch, the next time the Auto Freezer runs it'll pull a new copy of the repo and make a fresh branch from the latest version of master.

I'll give it a proper review after work.

techman83:
> If we decide not to freeze a particular idle mod, the script may keep re-submitting PRs to freeze it. Ideally we would create a way to suppress this.

Do we really need to worry about this? I say it's not worth adding logic for a rather edge case problem. If someone revives a mod, unfreezing it will inflate it straight away and update last indexed.

HebaruSan:
> Do we really need to worry about this?

Maybe, maybe not. I had a nagging sense at the back of my mind that I'd missed something, and that was an attempt to bring it to the foreground. If we make a hard assumption that we will always freeze these mods, then submitting a pull request is presenting a choice (merge or not) that isn't a choice at all, and we should just push to master instead.

HebaruSan:
Note to self, how to list the stale mods from the command line:

`wget -q -O - http://status.ksp-ckan.space/status/netkan.json | jq -rS '("2017-01-01T00:00:00Z"|fromdate) as $cutoff | to_entries | map(select(.value.last_indexed and (.value.last_indexed[0:19]+"Z"|fromdate)<$cutoff) | .key) | sort_by(ascii_downcase) | .[]'`

techman83:
> Maybe, maybe not. I had a nagging sense at the back of my mind that I'd missed something, and that was an attempt to bring it to the foreground. If we make a hard assumption that we will always freeze these mods, then submitting a pull request is presenting a choice (merge or not) that isn't a choice at all, and we should just push to master instead.

I saw it as more of a sanity check rather than a choice. Ie something went wrong and it tried to freeze all the Netkans and we didn't notice for a few days.

netkan/netkan/auto_freezer.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def mark_frozen_mods(self):
+        with ModStatus.batch_write() as batch:
+            logging.info('Marking frozen mods...')
+            for mod in ModStatus.scan(rate_limit=5):
+                if not mod.frozen and self._is_frozen(mod.ModIdentifier):
+                    logging.info('Marking frozen: %s', mod.ModIdentifier)
+                    mod.frozen = True
+                    batch.save(mod)
+            logging.info('Done!')
```
techman83: This seems like a bit of double handling, but rate limited DynamoDB writes are part of the free tier. We _could_ potentially do this with a webhook, but  I think doing this once a week is ok for now and will catch when we mark things as frozen by hand in the interim.

To note `batch_write` requires some special handling, but we're getting rate limiting on our writes as a consequence of our ModStatus scan.

netkan/netkan/auto_freezer.py
```diff
@@ -0,0 +1,88 @@
...
+        return not Path(self.netkan_repo.working_dir, 'NetKAN', f'{ident}.netkan').exists()
+
+    def _checkout_branch(self, name):
+        try:
+            self.netkan_repo.remotes.origin.fetch(name)
+        except git.GitCommandError:
+            logging.info('Unable to fetch %s', name)
+
+        (getattr(self.netkan_repo.heads, name, None)
+         or self.netkan_repo.create_head(name)).checkout()
```
techman83: I think we should probably build a function for branch checkouts, one with tests, that we can use as they can be right finicky. Checking out the branch is one of the first things we do, so if it blows up it'll get logged and the container will bail out.

PR #97, opened 2019-11-17
Add recover-status-timestamps command

HebaruSan:
Some mods have `null` timestamps, usually because they were only ever inflated by the legacy webhooks, which did not update the status file. This is inconvenient for changes that use these timestamps, such as #96.

Now a new `recover-status-timestamps` command is added that sets each null `ModStatus.last_indexed` to the timestamp of the newest commit in CKAN-meta for that module. This command is a one-off script and so it is not configured to run automatically; we would need to run it once on the server manually to backfill these timestamps.

### DIFF
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -161,6 +161,15 @@ def restore_status(filename):
 
 @click.command()
 @click.option(
+    '--ckanmeta-remote', envvar='CKANMETA_REMOTE',
+    help='Path/URL/SSH to Metadata Repo',
+)
+def recover_status_timestamps(ckanmeta_remote):
+    ModStatus.recover_timestamps(init_repo(ckanmeta_remote, '/tmp/CKAN-meta'))
+
+
+@click.command()
+@click.option(
     '--cluster', help='ECS Cluster running the service'
 )
 @click.option(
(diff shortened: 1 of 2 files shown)


DasSkelett:
I'm not happy with "cluttering" the cli.py with one-offs. But tbh, I don't have an alternative in mind either.
Maybe this could benefit from #92? We could put it at least in a separate file then.

HebaruSan:
Agreed, but we do already have a one-off in there (the job that copied the old `status.json` to `ModStatus`), so I think it's sound in principle:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/1eed50ce58a7688b2270397d0a3396e4c9c8603c/netkan/netkan/cli.py#L152-L159

techman83:
> One possible advantage of #92 here is that we could put normal commands and one-off commands in separate folders. `netkan/netkan/cli` vs `netkan/netkan/cli/scripts` e.g.

We could also use a different python entrypoint, say `admin_cli.py`

```python
netkan-admin=netkan.admin_cli:admin
```

techman83:
LGTM, I'll merge this and give it a run. I've taken a sanity copy of the status dump just in case.

PR #99, opened 2019-11-18
Timestamp recovery logging, timezones, required param

HebaruSan:
## Problem

#97 was merged and run, but we still have lots of missing `last_indexed` values.

## Cause

Unknown; it seemed to work fine on the test db. We need to know more about what it's doing.

## Changes

- Now we log an info message at the start and end
- Now we log an info message for each mod we check and save
- Now the `--ckanmeta-remote` parameter is marked as required because we can't run without it
- Now we convert timestamps to UTC before saving them

### DIFF
--- a/netkan/netkan/status.py
+++ b/netkan/netkan/status.py
@@ -2,7 +2,7 @@ import json
 import logging
 import os
 import time
-from datetime import datetime
+from datetime import datetime, timezone
 from pathlib import Path
 import boto3
 from dateutil.parser import parse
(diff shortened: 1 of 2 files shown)


netkan/netkan/status.py
```diff
@@ -108,7 +108,7 @@ def restore_status(cls, filename):
     @classmethod
     def last_indexed_from_git(cls, ckanmeta_repo, identifier):
         try:
-            return parse(ckanmeta_repo.git.log('--', identifier, format='%aI', max_count=1).split("\n")[0])
+            return parse(ckanmeta_repo.git.log('--', identifier, format='%aI', max_count=1).split("\n")[0]).astimezone(timezone.utc)
         except Exception as exc:  # pylint: disable=broad-except
```
techman83: As noted on @DasSkelett 's PR, the timezone stuff isn't required. PynamoDB's `UTCDateTimeAttribute` takes care of this. But as a once off task, this is NBD.

PR #100, opened 2019-11-20
Autofreezer Fixes for Indexing

techman83:
Noted an error trying to update the stack
```
Parameter ScheduleExpression is not valid. (Service: AmazonCloudWatchEvents; Status Code: 400; Error Code: ValidationException; Request ID: 31c198c9-8f8e-4129-a3c2-a5186729e7b7)
```
There is a commit that fixes this.

There is also a commit as for the current state of the stack. I've reverted the Change to the NetKAN repo constant and disabled the AutoFreezer task. 

Commits can be added here and I'll re-deploy tomorrow. Issue is that the tasks aren't initialising an SSH Key as HTTP doesn't require it. There was a secondary issue in that the bot has never written to the NetKAN repo and thus didn't have the permissions (I have rectified that).

Some minor adjustments will be required for the last PR to work properly.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,6 +31,7 @@ services:
       target: dev
     environment:
       NETKAN_REMOTE: ${NETKAN_METADATA_PATH}
+      SSH_KEY: ${CKAN_NETKAN_SSHKEY}
       AWS_DEFAULT_REGION: ${CKAN_AWS_DEFAULT_REGION}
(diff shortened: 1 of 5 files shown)


techman83:
Thanks for that @HebaruSan. I really need to get to refactoring the stacks/cli, there are some really nice python features we can lean on to reduce code duplication and protect us from missing things.

PR #101, opened 2019-11-23
Port version comparison logic from CKAN-core to metadata.py

DasSkelett:
## Motivation
See https://github.com/KSP-CKAN/CKAN/pull/2824, it is planned to implement an auto-epoch feature, which should kick in when an author screws up versioning again (numerous examples linked in the above PR).

Since the version comparison logic is fairly complex, it was first planned to use [CKAN-Core's implementation](https://github.com/KSP-CKAN/CKAN/blob/14ab1522e410c2bb0860d60b8986f625ec42a78e/Core/Versioning/ModuleVersion.cs#L191-L365) for that (via the command line interface f.e.), however spinning up mono multiple times per mod to find the highest version during indexing is rather resource heavy.
So here is the PR that ports the logic to the netkan package.

## Additions & Changes
The `version` attribute of a `Ckan` object is no longer a simple string, but an instance of `Ckan.Version`, a new sub-class.

The only "bigger" difference in the constructor is that it manually assigns a `0` as epoch, if none is given. This is because further down the epochs are checked for equality, and while C# handles that just well and returns `true` is both epochs are `null` or one is `null` and the other isn't and so on, Python really doesn't like `None`s and throws exceptions.
This is totally fine, because our [Spec says](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#epoch) that it assumes `0` if no epoch is given.

Else it is still based on the same regex, with just `?` swapped to `P`s for the named groups.

The comparison logic is as close to the C# one as possible. So this is definitely not pythonic.
The biggest problem was, that Python3 deprecated the `__cmp__` method, and the C# implementation is highly built on a similar concept (`-1` means 'less than', `0` means 'equal', `1` means 'greater than'), especially in the `stringComp()` and `numComp` methods.

I built around this by 'simulating' the `__cmp__` return values with this trick:
`_result = ((a > b)-(a < b))`
It's ugly but it works.

Furthermore the logic is not implemented as `__cmp__()` which would match C#'s `CompareTo()`, but as `__gt__()`. This resulted in some changes, like not returning a number indicating the relation of the versions, but simply `True` or `False`. The other possible relationships are all calculated from this.

I implemented the `Comparison` struct as a triple `(result, first_remainder, second_remainder)` which works really well!
I added the `else: return 0` in line 199+200, because the C# code makes use of `0` as default value for ints right there.

---

I also ported all the unit tests of `CKAN/Tests/Core/Versioning/Version.cs`, except two, the one testing construction of `UnmanagedModuleVersion` and the one for `ProvidesModuleVersion`.

---

Please go ahead and review carefully!

### DIFF
--- a/netkan/netkan/metadata.py
+++ b/netkan/netkan/metadata.py
@@ -1,5 +1,6 @@
 import json
 import re
+from functools import total_ordering
 from pathlib import Path
 from hashlib import sha1
 import uuid
@@ -83,6 +84,7 @@ class Ckan:
         else:
             self.contents = contents
         self._raw = json.loads(self.contents)
+        self.version = self.Version(self._raw.get('version'))
 
     def __getattr__(self, name):
         if name in self._raw:
@@ -93,7 +95,7 @@ class Ckan:
 
     @property
     def cache_prefix(self):
-        if not 'download' in self._raw:
+        if 'download' not in self._raw:
             return None
         return sha1(self.download.encode()).hexdigest().upper()[0:8]
 
@@ -104,7 +106,7 @@ class Ckan:
         return '{}-{}-{}.{}'.format(
             self.cache_prefix,
             self.identifier,
-            self.version.replace(':', '-'),
+            self.version.string.replace(':', '-'),
             self.MIME_TO_EXTENSION[self.download_content_type],
         )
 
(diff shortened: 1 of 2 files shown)


netkan/netkan/metadata.py
```diff
@@ -121,3 +122,164 @@ def licenses(self):
...
+            else:
+                # https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#epoch
+                # 0 assumed if no epoch supplied
+                self.epoch = 0
+            self.bare_version = match.group('version')
+            if self.bare_version is None:
+                raise Exception
+
+        def __eq__(self, other):
+            return not self < other and not self > other
```
HebaruSan: Could we compare `self.string == other.string` here? I think that would always give the right answer.

netkan/netkan/metadata.py
```diff
@@ -121,3 +122,164 @@ def licenses(self):
...
+
+        def __ne__(self, other):
+            return not self == other
+
+        def __lt__(self, other):
```
HebaruSan: The `@total_ordering` decoration from `functools` can apparently generate these stubs for us.

netkan/netkan/metadata.py
DasSkelett: This would break `test_issue1076()`

netkan/netkan/metadata.py
```diff
@@ -121,3 +122,164 @@ def licenses(self):
             return lic
         else:
             return [lic]
+
+    class Version:
+
+        Pattern = re.compile("^(?:(?P<epoch>[0-9]+):)?(?P<version>.*)$")
```
HebaruSan:
```suggestion
        PATTERN = re.compile("^(?:(?P<epoch>[0-9]+):)?(?P<version>.*)$")
```

I think we're going with the convention that class constants are in ALL CAPS.

netkan/netkan/metadata.py
HebaruSan: Ahh, I see.

netkan/netkan/metadata.py
```diff
@@ -121,3 +122,164 @@ def licenses(self):
...
+                # if they are the exhausted parts will return zero.
+                (result, first_remainder, second_remainder) = _number_compare(first_remainder, second_remainder)
+
+                # Again, return difference if found.
+                if result != 0:
+                    return result > 0
+
+            # Oh, we've run out of one or both strings.
+            if len(first_remainder) == 0:
+                if len(second_remainder) == 0:
```
HebaruSan: This test looks redundant. If it's true, we'll return `False` in the next line. Otherwise we'll return `False on line 281. Either way we always return false.

netkan/netkan/metadata.py
DasSkelett: It works fine for everything except `__eq__`, so I'll keep that one.

DasSkelett:
Amended to the commit to address the comments. `@total_ordering` has problems with `__eq__`, even basic equal tests with versions with the same string won't succeed, so I kept that one. The other generated rich comparison functions appear to work fine.

netkan/netkan/metadata.py
HebaruSan: Yup. :+1: 

> The class must define one of __lt__(), __le__(), __gt__(), or __ge__(). In addition, the class should supply an __eq__() method.

https://docs.python.org/2/library/functools.html

netkan/netkan/metadata.py
```diff
@@ -83,6 +84,7 @@ def __init__(self, filename=None, contents=None):
         else:
             self.contents = contents
         self._raw = json.loads(self.contents)
+        self.version = self.Version(self._raw.get('version'))
```
techman83: I this supposed to be
```python
self.version = Version(self._raw.get('version'))
```

HebaruSan: `Version` is defined within the body of `Ckan`, so this is shorthand for `Ckan.Version`, kind of like a class constant, I thought.

techman83: Oh so it is, missed the indenting in GitHub. That's an interesting approach, makes sense that is closely tied to Ckan. But if we ever want to use it outside the Ckan class we might want to make it a top level.

techman83:
Lots of tests, looks great! Merging with thanks!

netkan/netkan/metadata.py
DasSkelett: You can still use it like this: `some_version = Ckan.Version('somestring')`, without instantiating a `Ckan` object, like I did in the tests.
But yes, the option is there to move it out of `Ckan` if there is a need for it.

techman83: You can, and that's fine for testing. But I think we would make it a top level class if we needed it beyond that.

PR #102, opened 2019-11-25
Send HighestVersion attribute to Inflator

### LINKED ISSUE #7: [Feature] Support HighestVersion attribute for auto epoching
This is an outgrowth of https://github.com/KSP-CKAN/CKAN/issues/2789#issuecomment-507468298. I assume that the initial functionality of the queue-based bot will be identical to NetKAN-bot, so ideas for further improvements should be submitted here for better tracking.

## Motivation

See KSP-CKAN/CKAN#2824. We use epochs to coerce author-created versions into a sensible ordering, but this requires a lot of manual intervention and could potentially be automated.

## Suggestion

Before submitting a netkan to be inflated, scan the existing versions of that module and compare their versions to find the maximum. Send this in the `HighestVersion` attribute of the queue 
message. Proof of concept code:

https://gist.github.com/HebaruSan/1fc031cbd97a9a1db2c7bf3c46e9894f

This runs slowly because it executes mono and `ckan.exe` once per version (but it does output the right answer). To make it acceptably fast, we would need to rewrite the version comparison logic in Python:

https://github.com/KSP-CKAN/CKAN/blob/14ab1522e410c2bb0860d60b8986f625ec42a78e/Core/Versioning/ModuleVersion.cs#L191-L365

Or we could write a new `ckan.exe max_version a b c d` command, etc.

HebaruSan:
## Motivation

Now that #101 has given us the ability to compare module versions in Python the same as in CKAN-Core, this is the Scheduler + Webhook part of KSP-CKAN/CKAN#2824, which aims to reduce the need for manual maintenance of `x_netkan_epoch` properties.

## Changes

A new `CkanGroup` class represents all of a given module's .ckan files in a CKAN-meta repo. This class provides a `highest_version` function that parses the .ckan files and compares their `version` properties to find the highest one, and returns it.

The `Ckan.sqs_message` function now accepts a `CkanGroup` parameter and passes it to a new `sqs_message_attribs` function to generate the `MessageAttributes` property of the message. Naturally, this uses `CkanGroup.highest_version` to set `HighestVersion`.

If `x_netkan_allow_out_of_order` is set to `true` in a netkan, this means auto-epoching is turned off, so the `HighestVersion` property will not be set and its .ckan files will not be parsed.

The Scheduler now checks out CKAN-meta and uses it to pass `CkanGroup` objects to `Ckan.sqs_message`. The Webhooks also pass `CkanGroup` objects to `Ckan.sqs_message`.

Fixes #7.


netkan/netkan/metadata.py
```diff
@@ -59,12 +59,19 @@ def hook_only(self):
...
         return {
-            'Id': self.identifier,
-            'MessageBody': self.contents,
-            'MessageGroupId': '1',
+            'Id':                     self.identifier,
+            'MessageBody':            self.contents,
+            'MessageGroupId':         '1',
             'MessageDeduplicationId': uuid.uuid4().hex,
+            'MessageAttributes':      self.sqs_message_attribs(ckan_group),
         }
```
techman83: PEP8 Nitpick -> https://www.python.org/dev/peps/pep-0008/#pet-peeves

netkan/netkan/metadata.py
```diff
@@ -278,3 +287,16 @@ def _number_compare(v1, v2) -> (int, str, str):
...
+
+    def __init__(self, ckan_meta_path, identifier):
+        self.path = Path(ckan_meta_path, identifier)
+
+    def highest_version(self):
+        ckans = (Ckan(p) for p in self.path.glob('**/*.ckan'))
+        highest = max(ckans, default=None, key=lambda ck: ck.version)
+        if highest:
+            return highest.version
+        return None
```
techman83: This is where a doc string might be handy, explaining what a CkanGroup actually is. It is a group of Ckans, but it's tighter in scope than just a group. Naming is hard! Something like

```python
class CkanGroup:

    """A collection of ckans grouped by their identifier"""

   def ....
```

netkan/netkan/scheduler.py
```diff
@@ -42,7 +43,8 @@ def _in_group(self, netkan):
             return self.nonhooks_group
 
     def schedule_all_netkans(self):
-        messages = (nk.sqs_message() for nk in self.netkans() if self._in_group(nk))
+        messages = (nk.sqs_message(CkanGroup(self.ckan_meta_path, nk.identifier))
+                    for nk in self.netkans() if self._in_group(nk))
         for batch in sqs_batch_entries(messages):
```
techman83: I don't think it's a big deal, but do we know what the performance hit of scheduling like this?

netkan/netkan/metadata.py
techman83: We're already testing the version stuff directly, but it'd be good to have some directly for this function. Save us surprises if we decide to refactor down the track.

netkan/netkan/scheduler.py
HebaruSan: I couldn't think of a good way to measure that. Ballparking, it adds disk reads and JSON parsing, to the tune of several times what's currently done (since each module has one .netkan and multiple .ckans). The scheduler might take, say, 10x as much time to run, but I'm pretty sure it's not the current bottleneck, either.

techman83: Admittedly my dev machine at work is pretty beefy, single threaded performance isn't going to be too different
```python
In [18]: from netkan.metadata import Ckan                                                                                                                                                                                                                                                                                    

In [19]: from pathlib import Path                                                                                                                                                                                                                                                                                            

In [20]: from timeit import default_timer as timer                                                                                                                                                                                                                                                                           

In [21]: def load_all_ckans(identifiers): 
    ...:     start = timer() 
    ...:     for identifier in identifiers: 
    ...:         ckans = (Ckan(p) for p in identifier.glob('**/*.ckan')) 
    ...:         highest = max(ckans, default=None, key=lambda ck: ck.version) 
    ...:         highest.identifier if highest else None 
    ...:     end = timer() 
    ...:     print(end - start) 
    ...:                                                                                                                                                                                                                                                                                                                     

In [22]: load_all_ckans(Path('../CKAN-meta/').iterdir())                                                                                                                                                                                                                                                                     
1.4594180890126154

In [23]: load_all_ckans(Path('../CKAN-meta/').iterdir())                                                                                                                                                                                                                                                                     
1.4571303109405562
```
And that's all of them, not just the ones we'll be scheduling.  So I think we'll be fine and dandy for a synchronous scheduled task.

HebaruSan:
The Indexer will not need any changes for this; it's solely a Scheduler+Inflator change.

netkan/netkan/metadata.py
HebaruSan: Do any of those apply to this case? Doesn't look like it:

> Immediately inside parentheses, brackets or braces.
> Between a trailing comma and a following close parenthesis.
> Immediately before a comma, semicolon, or colon:
> Immediately before the open parenthesis that starts the argument list of a function call:
> Immediately before the open parenthesis that starts an indexing or slicing:
> More than one space around an assignment (or other) operator to align it with another.

techman83: This, even if it doesn't specifically mention dictionaries.
> More than one space around an assignment (or other) operator to align it with another.

techman83:
Yep, after refreshing myself I realised we could re-utilise the staging logic without any changes to the indexer. Which is rather neat.

netkan/netkan/metadata.py
HebaruSan: OK.

PR #105, opened 2019-11-29
Allow gunicorns after 20.0.0

### LINKED ISSUE #85: [Bug] Webhooks fail to start under gunicorn 20.0.0
Background
----------
After releasing the latest batch of changes the webhooks failed to start. Not being part of the python code, rather gunicorn it didn't send the logs to discord.

Worked around it by locking below 20.0.0 for now.

Problem
-------
```bash
netkan@bef8cb5997ae:~/workspace$ gunicorn -b 0.0.0.0:5000 --access-logfile - "netkan.webhooks:create_app()"
[2019-11-11 03:31:32 +0000] [239] [INFO] Starting gunicorn 20.0.0
[2019-11-11 03:31:32 +0000] [239] [INFO] Listening at: http://0.0.0.0:5000 (239)
[2019-11-11 03:31:32 +0000] [239] [INFO] Using worker: sync
[2019-11-11 03:31:32 +0000] [243] [INFO] Booting worker with pid: 243
Failed to find application object 'create_app()' in 'netkan.webhooks'
[2019-11-11 03:31:33 +0000] [243] [INFO] Worker exiting (pid: 243)
[2019-11-11 03:31:33 +0000] [239] [INFO] Shutting down: Master
[2019-11-11 03:31:33 +0000] [239] [INFO] Reason: App failed to load.
```

HebaruSan:
## Problem

gunicorn is currently pinned to version 19.9.0.

## Cause

The 20.0.0 release broke support for the factory pattern, which we use: benoitc/gunicorn#2159

## Changes

Version 20.0.1 fixed the issue (20.0.2 includes notes for 20.0.1):

https://github.com/benoitc/gunicorn/releases/tag/20.0.2

> fixed support of applications loaded from a factory function

I tried it out (with 20.0.3) and it seems to work, so now the pin is changed to just exclude 20.0.0.

Fixes #85.

### DIFF
--- a/netkan/setup.py
+++ b/netkan/setup.py
@@ -19,7 +19,7 @@ setup(
         'requests',
         'flask',
         'internetarchive',
-        'gunicorn>=19.9,<20.0.0',
+        'gunicorn>=19.9,!=20.0.0',
         'discord',
         'PyGithub',
     ],


netkan/setup.py
```diff
@@ -19,7 +19,7 @@
         'requests',
         'flask',
         'internetarchive',
-        'gunicorn>=19.9,<20.0.0',
+        'gunicorn>=19.9,!=20.0.0',
         'discord',
```
techman83: We could probably just require >20.0.0, but it doesn't really make much difference for us.

PR #107, opened 2019-12-01
Revert unintentional restoration of debug param to download counter

HebaruSan:
## Problem

```
Uncaught exception:
    sys.exit(netkan())
    rv = self.invoke(ctx)
TypeError: download_counter() missing 1 required positional argument: 'debug'
```

## Cause

An unused `debug`param was added to `download_counter` without Click definitions in #102, probably due to conflict resolution with #86.

## Changes

Now the parameter is removed.

### DIFF
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -256,7 +256,7 @@ def clean_cache(days):
     '--key', envvar='SSH_KEY', required=True,
     help='SSH key for accessing repositories',
 )
-def download_counter(netkan_remote, ckanmeta_remote, token, key, debug):
+def download_counter(netkan_remote, ckanmeta_remote, token, key):
     init_ssh(key, '/home/netkan/.ssh')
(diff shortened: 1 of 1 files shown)


techman83:
I really want to refactor all this! So many places to get caught.

DasSkelett:
I think unit tests calling the click commands themselves would already catch a lot of those quirks, but this would need access to an SQS system + database.
There's https://github.com/localstack/localstack for running SQS locally for example, but this looks fairly heavy.

techman83:
That's true. But we're doing a lot of repeated work, that we have to remember to add for each function. Writing a custom click decorator + class would go a long way to addressing that. We could have a 'Netkan' command class, that takes a list of mandatory options and the rest would just be available if our caller wanted them. Things like repos/keys would have getters that would initialise them when required etc. We could move our logging initialisation out of the group into the command and at that point potentially break up the CLI a bit, not so much for saving from mistakes, rather scoping what each of the commands load into memory.

PR #108, opened 2019-12-15
Pull both repos before handling webhooks

HebaruSan:
## Problem

screenshot

This module should have been auto-epoched because `1.6_alpha` came out before `0.17_Alpha`, but it wasn't.

## Cause

The Webhooks have a CKAN-meta repo cloned so they can find the highest versions of mods to be inflated (see #102), but they never update it. This mod was indexed after the latest time the Infra was restarted, so the `1.6_alpha` version was not available, and `0.17_Alpha` did not get compared to it.

## Changes

Now a new `netkan.common.pull_all` function accepts a list of repos to pull, and the Webhooks put all of their repos into `current_app.config['repos']` for ease of access. Then in the places where we currently pull the NetKAN repo, we instead pull all of the repos in that list. This will ensure that CKAN-meta is up to date before the Webhook submits a mod for inflation.

### DIFF
--- a/netkan/netkan/common.py
+++ b/netkan/netkan/common.py
@@ -14,3 +14,8 @@ def sqs_batch_entries(messages, batch_size=10):
             batch = []
     if len(batch) > 0:
         yield batch
+
+
+def pull_all(repos):
+    for r in repos:
+        r.remotes.origin.pull('master', strategy_option='theirs')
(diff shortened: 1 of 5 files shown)


techman83:
Yes, I've been meaning to raise an issue about this. We're raising a lot of autoepoch PRs currently due to that not being updated. This should do the trick!

PR #112, opened 2019-12-16
Check Volume Stats before Scheduling

techman83:
Over the weekend we had a situation where there was a bulk update to the metadata. Commits seem to do an awful lot of IO operations, which normally isn't a big deal. But with all the changes to the metadata recently, it slowly pushed down the burnt balance to zero, at which point the system DoS'd itself and was never able to catch up.

This code probably needs a bit of a cleanup, but it does what it says on the tin (I had to fake a few things in dev to get it to access prod):
```
In [4]: NetkanScheduler('.','.','TestyMcTestFace').can_schedule(50,False)                                                                                                                 
[2019-12-16 13:00:47,018] [INFO    ] Found credentials in environment variables.
288.0 <--- Print of CPU Credits
31.4 <--- Print of volume burst credit percentage
```
It might do us well to check the outbound queue as well. But as a general rule the indexer is not the bottle neck as most inflations don't involve metadata changes (lucky to get a handful per run) and thus is normally processing them at a much faster rate than the inflator can inflate them.

### DIFF
--- a/netkan/netkan/scheduler.py
+++ b/netkan/netkan/scheduler.py
@@ -10,7 +10,8 @@ from .common import sqs_batch_entries
 
 class NetkanScheduler:
 
-    def __init__(self, path, ckan_meta_path, queue, base='NetKAN/', nonhooks_group=False, webhooks_group=False):
+    def __init__(self, path, ckan_meta_path, queue,
+                 base='NetKAN/', nonhooks_group=False, webhooks_group=False):
         self.path = Path(path, base)
         self.nonhooks_group = nonhooks_group
         self.webhooks_group = webhooks_group
@@ -48,6 +49,52 @@ class NetkanScheduler:
         for batch in sqs_batch_entries(messages):
             self.client.send_message_batch(**self.sqs_batch_attrs(batch))
 
+    def cpu_credits(self, cloudwatch, instance_id, start, end):
+        stats = cloudwatch.get_metric_statistics(
+            Dimensions=[{'Name': 'InstanceId', 'Value': instance_id}],
+            MetricName='CPUCreditBalance',
+            Namespace='AWS/EC2',
+            StartTime=start.strftime("%Y-%m-%dT%H:%MZ"),
+            EndTime=end.strftime("%Y-%m-%dT%H:%MZ"),
+            Period=10,
+            Statistics=['Average'],
+        )
+        # A pass consumes around 40 credits, with an accrue rate of 24/hr.
+        # So running every 2 hours should see using just a touch less than
+        # we gain in that time period.
+        creds = 0
+        try:
+            creds = stats['Datapoints'][0]['Average']
+        except IndexError:
+            logging.error("Couldn't acquire CPU Credit Stats")
+        return int(creds)
+
+    def volume_credits(self, cloudwatch, instance_id, start, end):
+        client = boto3.client('ec2')
+        response = client.describe_volumes(
+            Filters=[{
+                'Name': 'attachment.instance-id',
+                'Values': [instance_id]
+            }]
+        )
+        # If we add a second gp2 volume, this may break
+        volume = list(filter(lambda x: x['VolumeType'] == 'gp2', response['Volumes']))[0]
+        volume_id = volume['Attachments'][0]['VolumeId']
+        stats = cloudwatch.get_metric_statistics(
+            Dimensions=[{'Name': 'VolumeId', 'Value': 'vol-02cdb3dfd4b2a69f9'}],
+            MetricName='BurstBalance',
+            Namespace='AWS/EBS',
+            StartTime=start.strftime("%Y-%m-%dT%H:%MZ"),
+            EndTime=end.strftime("%Y-%m-%dT%H:%MZ"),
+            Period=10, Statistics=['Average'],
+        )
+        creds = 0
+        try:
+            creds = stats['Datapoints'][0]['Average']
+        except IndexError:
+            logging.error("Couldn't acquire Volume Credit Stats")
+        return int(creds)
+
     def can_schedule(self, max_queued, dev=False, min_credits=200):
         if not dev:
             end = datetime.datetime.utcnow()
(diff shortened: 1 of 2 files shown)


netkan/netkan/scheduler.py
```diff
@@ -57,26 +104,21 @@ def can_schedule(self, max_queued, dev=False, min_credits=200):
...
+                logging.info(
+                    "Run skipped, below cpu credit target (Current Avg: %s)", cpu_credits
+                )
+                return False
+
+            # Volume Burst balance measured in a percentage of 5.4million credits. Credits are
+            # accrued at a rate of 3 per GB, per second. If we are are down to 30 percent of
+            # our max, something has gone wrong and we should not queue any more inflations.
+            vol_credits = self.volume_credits(cloudwatch, instance_id, start, end)
+            if vol_credits < 30:
```
DasSkelett: The `min_credits` for the CPU are supplied via an argument. Should we consider the same for the volume credits? I know it's a percentage whereas the CPU credits are a total amount, so when we're upgrading the effect is not so severe, but it might still make sense.

techman83: Maybe, but this is more of a fall safe, rather than something we expect to hit regularly. Maybe it should log error, so we know when this has happened.

HebaruSan:
Thinking about how to lighten the load on the Indexer, if creating a commit is expensive, what if we added all changed modules to one commit per batch, rather than one per module? In a normal pass we'd be unlikely to have more than one changed module in a batch, but for mass-commit changes, it would decrease the number of commits by 90%.

netkan/netkan/scheduler.py
```diff
@@ -48,6 +49,52 @@ def schedule_all_netkans(self):
...
+        stats = cloudwatch.get_metric_statistics(
+            Dimensions=[{'Name': 'InstanceId', 'Value': instance_id}],
+            MetricName='CPUCreditBalance',
+            Namespace='AWS/EC2',
+            StartTime=start.strftime("%Y-%m-%dT%H:%MZ"),
+            EndTime=end.strftime("%Y-%m-%dT%H:%MZ"),
+            Period=10,
+            Statistics=['Average'],
+        )
+        # A pass consumes around 40 credits, with an accrue rate of 24/hr.
```
HebaruSan: Is this estimate (40 cred per pass) still up to date with current performance after KSP-CKAN/CKAN#2940 ? Not crucial to update it, but if someone does need the info, it'd be nice to have it accurate.

netkan/netkan/scheduler.py
```diff
@@ -48,6 +49,52 @@ def schedule_all_netkans(self):
...
+            Filters=[{
+                'Name': 'attachment.instance-id',
+                'Values': [instance_id]
+            }]
+        )
+        # If we add a second gp2 volume, this may break
+        volume = list(filter(lambda x: x['VolumeType'] == 'gp2', response['Volumes']))[0]
+        volume_id = volume['Attachments'][0]['VolumeId']
+        stats = cloudwatch.get_metric_statistics(
+            Dimensions=[{'Name': 'VolumeId', 'Value': 'vol-02cdb3dfd4b2a69f9'}],
```
HebaruSan: For the other metric we retrieve the id, it looks like here we've hard coded it. Is there a way to retrieve it instead?

netkan/netkan/scheduler.py
```diff
@@ -48,6 +49,52 @@ def schedule_all_netkans(self):
...
+        # we gain in that time period.
+        creds = 0
+        try:
+            creds = stats['Datapoints'][0]['Average']
+        except IndexError:
+            logging.error("Couldn't acquire CPU Credit Stats")
+        return int(creds)
+
+    def volume_credits(self, cloudwatch, instance_id, start, end):
+        client = boto3.client('ec2')
```
HebaruSan: I think we could use `self.client` instead of getting a new one here.

netkan/netkan/scheduler.py
techman83: I've made some adjustments here. Lemme know what you think :slightly_smiling_face:

netkan/netkan/scheduler.py
techman83: Nope, I made a mistake when testing the code :grimacing:

netkan/netkan/scheduler.py
techman83: Because we only use the ec2 client once. We could use self.cloudwatch instead of passing it in though.

techman83: Actually going to leave that for a later refactor. It's kinda hard to test without being on the instance, without doing a lot more work.

netkan/netkan/scheduler.py
HebaruSan: Looks like a helpful update! Thanks!

techman83:
> Thinking about how to lighten the load on the Indexer, if creating a commit is expensive, what if we added all changed modules to one commit per batch, rather than one per module? In a normal pass we'd be unlikely to have more than one changed module in a batch, but for mass-commit changes, it would decrease the number of commits by 90%.

We used to do this, but in general this isn't normally an issue. I think for the rare occasion a change to the inflator causes a change to the way the files a laid out, skipping a run or two while the credits build up is acceptable. It's likely not the commits themselves, rather the push/pull. Changing this would add a lot of complexity that I don't think is necessary right now.

I think we should look at _why_ our commits/pulls/pushes involve a large amount of IO operations. My hunch is the deep clone and as for the majority of our operations we don't care about the history, so we should switch back to a shallow clone (can be discussed in another issue/pr).

For now this should avoid us hitting the issue we had over the weekend. If there is nothing else required, I'll merge it after lunch.

HebaruSan:
> My hunch is the deep clone

I'm skeptical of this. It's an easy scapegoat to blame, but I doubt that git's data model slows down as its history grows, because that would be a poor design, and git is generally highly respected.

Shallow clones are faster **for cloning**, but I think that's the extent of the impact.

techman83:
> I'm skeptical of this. It's an easy scapegoat to blame, but I doubt that git's data model slows down as its history grows, because that would be a poor design, and git is generally highly respected.

Git yes, but I'm not 100% about the interactions with gitpython. It does pull a lot into memory, happy to be wrong though. Easy enough to test and there is nothing gained by having a deep clone operationally.

PR #114, opened 2019-12-19
Set up Discord notifications earlier in Webhooks

HebaruSan:
## Problem

The NetKAN inflation Webhooks are failing currently. GitHub receives:

```html
<html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx/1.17.6</center>
</body>
</html>
```

## Cause

Researching that error, it seems it can happen if the nginx proxy is unable to connect to the upstream server. This would happen if the Webhooks crashed somehow. No stack traces have been printed to Discord, so whatever the problem is, it must be happening before the Discord hook is initialized.

This suggests there was a problem with #108, which edited the initialization function for the Webhooks and was merged around the time this started happening (Dec 16). Unfortunately I don't see any problems with that change upon auditing the code, so we'll need more information to devise a fix.

## Changes

Now the first thing the Webhooks do is set up the Discord integration. This way we will be notified if there's an exception during the rest of the initialization. We also now set `sys.excepthook` to catch uncaught exceptions, since otherwise I think these might not be reported.

### DIFF
--- a/netkan/netkan/webhooks/__init__.py
+++ b/netkan/netkan/webhooks/__init__.py
@@ -1,9 +1,10 @@
 import os
+import sys
 import boto3
 from flask import Flask
 
 from ..utils import init_repo, init_ssh
-from ..notifications import setup_log_handler
+from ..notifications import setup_log_handler, catch_all
 from .errors import errors
 from .inflate import inflate
 from .spacedock_inflate import spacedock_inflate
@@ -12,7 +13,8 @@ from .github_inflate import github_inflate
 
 def create_app():
     # Set up Discord logger so we can see errors
-    setup_log_handler()
+    if setup_log_handler():
+        sys.excepthook = catch_all
 
     app = Flask(__name__)
 
--- a/netkan/netkan/webhooks/__init__.py
+++ b/netkan/netkan/webhooks/__init__.py
@@ -11,6 +11,9 @@ from .github_inflate import github_inflate
 
 
 def create_app():
+    # Set up Discord logger so we can see errors
+    setup_log_handler()
+
     app = Flask(__name__)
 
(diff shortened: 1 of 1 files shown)


techman83:
LGTM, not entirely sure what's going on either as the logs don't float up to gunicorn (which is not uncommon)

```
[2019-12-17 05:34:25 +0000] [1] [CRITICAL] WORKER TIMEOUT (pid:8)
[2019-12-17 05:34:25 +0000] [8] [INFO] Worker exiting (pid: 8)
[2019-12-17 05:34:25 +0000] [31] [INFO] Booting worker with pid: 31
172.17.0.7 - - [17/Dec/2019:12:42:11 +0000] "POST /inflate HTTP/1.0" 204 0 "-" "python-requests/2.22.0"
[2019-12-18 15:26:08 +0000] [94] [INFO] Booting worker with pid: 94
[2019-12-18 20:35:08 +0000] [118] [INFO] Booting worker with pid: 118
[2019-12-18 20:35:19 +0000] [140] [INFO] Booting worker with pid: 140
[2019-12-19 14:55:57 +0000] [165] [INFO] Booting worker with pid: 165
[2019-12-19 19:49:22 +0000] [197] [INFO] Booting worker with pid: 197
[2019-12-20 00:11:07 +0000] [222] [INFO] Booting worker with pid: 222
[2019-12-20 01:26:57 +0000] [244] [INFO] Booting worker with pid: 244
```

HebaruSan:
Hmm, still doesn't work and no errors in Discord. Maybe the problem is internal to nginx?

techman83:
It hasn't quite finished redeploying yet. Need to add a notification for that!

PR #115, opened 2019-12-28
Set More Conservative Burst Limit

techman83:
We hit a situation where our burst credits dropped below a point where the indexer could recover from it. This PR sets a more conservative limit and also adds a quick fix to the logging line that was possibly preventing the log to show up in our notifications.


techman83:
Yeah, for sure. Hopefully buys us a little time for now though.

PR #116, opened 2020-01-14
Change absolute pathes to cache and .ssh to relative ones

DasSkelett:
## Problem
In #76 I added a test for `cli.clean_cache()`. Because `clean_cache()` uses an absolute path to the cache (`/home/netkan/ckan_cache`) this thest failed when running the tests outside a Docker container.

## Changes
All paths beginning with `/home/netkan` have been replaced with `Path.home()`.
Those were only pointing either to `/home/netkan/.ssh/` or `/home/netkan/ckan_cache/`.
`clean_cache()` takes another argument, `--cache` which is the path to the cache directory it should use. It defaults to `str(Path.home()) + '/ckan_cache/`.

The `test_clean_cache()` creates a `test_cache` folder in it's parent directory, which is used to run the test (better than running them in `/home/{user}/ckan_cache` which will be used for production later on, and creating folders on the host machine of devs running the tests).

The Dockerfile needed an update to `chown -R netkan:netkan` the `/netkan` folder in which the `test_cache` folder is created during the tests in the container.

### DIFF
--- a/netkan/Dockerfile
+++ b/netkan/Dockerfile
@@ -1,7 +1,6 @@
 FROM python:3.7 as base
 RUN useradd -ms /bin/bash netkan
 ADD . /netkan
-RUN chown -R netkan:netkan /netkan
 WORKDIR /netkan
 USER netkan
 RUN pip install --user . --no-warn-script-location
@@ -22,7 +21,6 @@ CMD ["--help"]
 FROM production as dev
 USER root
 ADD . /netkan
-RUN chown -R netkan:netkan /netkan
 ADD run_dev.sh /usr/local/bin/
 USER netkan
 RUN pip install --user /netkan/.[development]
--- a/netkan/Dockerfile
+++ b/netkan/Dockerfile
@@ -1,6 +1,7 @@
 FROM python:3.7 as base
 RUN useradd -ms /bin/bash netkan
 ADD . /netkan
+RUN chown -R netkan:netkan /netkan
 WORKDIR /netkan
 USER netkan
 RUN pip install --user . --no-warn-script-location
@@ -21,6 +22,7 @@ CMD ["--help"]
 FROM production as dev
 USER root
 ADD . /netkan
+RUN chown -R netkan:netkan /netkan
 ADD run_dev.sh /usr/local/bin/
 USER netkan
 RUN pip install --user /netkan/.[development]
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -60,7 +60,7 @@ def netkan(debug):
     help='SSH key for accessing repositories',
 )
 def indexer(queue, ckanmeta_remote, token, repo, user, key, timeout):
-    init_ssh(key,  str(Path.home()) + '/.ssh')
+    init_ssh(key,  Path(Path.home(), '.ssh'))
     ckan_meta = init_repo(ckanmeta_remote, '/tmp/CKAN-meta')
 
     github_pr = GitHubPR(token, repo, user)
@@ -120,7 +120,7 @@ def indexer(queue, ckanmeta_remote, token, repo, user, key, timeout):
     help='Only schedule if we have at least this many credits remaining',
 )
 def scheduler(queue, netkan_remote, ckanmeta_remote, key, max_queued, dev, group, min_credits):
-    init_ssh(key, str(Path.home()) + '/.ssh')
+    init_ssh(key, Path(Path.home(), '.ssh'))
     sched = NetkanScheduler(
         Path('/tmp/NetKAN'), Path('/tmp/CKAN-meta'), queue,
         nonhooks_group=(group == 'all' or group == 'nonhooks'),
@@ -228,7 +228,8 @@ def redeploy_service(cluster, service_name):
 )
 @click.option(
     '--cache', envvar='NETKAN_CACHE', default=str(Path.home()) + '/ckan_cache/',
-    help='Absolute path to the mod download cache',
+    type=click.Path(exists=True, writable=True),
+    help='Absolute path to the mod download cache'
 )
 def clean_cache(days, cache):
     older_than = (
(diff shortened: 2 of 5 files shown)


netkan/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM python:3.7 as base
 RUN useradd -ms /bin/bash netkan
 ADD . /netkan
+RUN chown -R netkan:netkan /netkan
 WORKDIR /netkan
```
techman83: This isn't necessary, in production this path ceases to exist and in dev we'll either use the volume mount under home (via docker-compose) or the workspace (from within vs-code).

netkan/Dockerfile
```diff
@@ -21,6 +22,7 @@ CMD ["--help"]
 FROM production as dev
 USER root
 ADD . /netkan
+RUN chown -R netkan:netkan /netkan
 ADD run_dev.sh /usr/local/bin/
```
techman83: As above.

netkan/netkan/cli.py
```diff
@@ -60,7 +60,7 @@ def netkan(debug):
     help='SSH key for accessing repositories',
 )
 def indexer(queue, ckanmeta_remote, token, repo, user, key, timeout):
-    init_ssh(key, '/home/netkan/.ssh')
+    init_ssh(key,  str(Path.home()) + '/.ssh')
     ckan_meta = init_repo(ckanmeta_remote, '/tmp/CKAN-meta')
```
techman83: `Path` can concatenate things itself and we can pass a `Path` to that function.

```python
init_ssh(key, Path(Path.home(), '.ssh'))
```

netkan/netkan/cli.py
```diff
@@ -120,7 +120,7 @@ def indexer(queue, ckanmeta_remote, token, repo, user, key, timeout):
     help='Only schedule if we have at least this many credits remaining',
 )
 def scheduler(queue, netkan_remote, ckanmeta_remote, key, max_queued, dev, group, min_credits):
-    init_ssh(key, '/home/netkan/.ssh')
+    init_ssh(key, str(Path.home()) + '/.ssh')
     sched = NetkanScheduler(
```
techman83: As above.

netkan/netkan/cli.py
```diff
@@ -226,12 +226,16 @@ def redeploy_service(cluster, service_name):
 @click.option(
     '--days', help='Purge items older than X from cache',
 )
-def clean_cache(days):
+@click.option(
+    '--cache', envvar='NETKAN_CACHE', default=str(Path.home()) + '/ckan_cache/',
+    help='Absolute path to the mod download cache',
+)
+def clean_cache(days, cache):
     older_than = (
```
techman83: Neat! I like this, same comment as above. If you want to play around with setting a type here, that could be interesting (but I will happily merge without it).

https://click.palletsprojects.com/en/7.x/api/?highlight=click%20path#click.Path

example use here:

https://click.palletsprojects.com/en/7.x/options/?highlight=click%20path#multiple-values-from-environment-values

netkan/tests/cli.py
```diff
@@ -45,7 +45,7 @@ def tearDown(self):
 
     def test_clean_all(self):
 
-        result = self.runner.invoke(clean_cache, ['--days', '42'])
+        result = self.runner.invoke(clean_cache, ['--days', '42', '--cache', str(self.cache_path)])
```
techman83: Do we definitely need to call string here? `Path` _should_ do the RightThing&trade; in most circumstances.

techman83:
> The Dockerfile needed an update to chown -R netkan:netkan the /netkan folder in which the test_cache folder is created during the tests in the container.

I missed reading this above regarding my comments. If we're creating temporary directories for things I'm in favour of using the tools to do that.

https://docs.python.org/3.7/library/tempfile.html#tempfile.TemporaryDirectory

There are examples in the other tests of using them this way. But happy to change my mind on the chown if it makes more sense or is too awkward.

netkan/netkan/cli.py
DasSkelett: Now passing Path objects to `init_ssh()`

netkan/netkan/cli.py
DasSkelett: Cool! Probably only useful for unit tests right now, as we don't run an integration test with the values we use in production, so we can't catch errors in the `prod-stack.yml`. But better than nothing.

netkan/tests/cli.py
DasSkelett: I think this is now irrelevant with the change to `TemporaryDirectory`.
But I don't think it would have converted to a string implicitly, because this test doesn't:
```python
>>> print(['1', '2', Path('3')])
['1', '2', PosixPath('3')]
```
Or do you mean click would call `str()` on everything non-string when running `invoke()` down the line?

netkan/Dockerfile
DasSkelett: Now obsolete thanks to `TemporaryDirectory()`

netkan/netkan/utils.py
```diff
@@ -14,12 +14,11 @@ def init_repo(metadata, path):
     return repo
 
 
-def init_ssh(key, path):
+def init_ssh(key, key_path: Path):
     if not key:
```
techman83: This is super neat, lets us complain loudly if we don't get a Path object.

netkan/tests/cli.py
techman83: I agree, `TemporaryDirectory.name` will be a string. As for your example calling  the `__str__` method on a list won't be calling the `__str__` methods on the members of that list, rather the `__repr__` method.
```
In [4]: for p in ['1', '2', Path('3')]: 
   ...:     print(p) 
   ...:                                                                                                                                                                                                                                                                                                                      
1
2
3
```
So it will probably depend on what happens inside of invoke.

DasSkelett: That's interesting, thanks!

netkan/netkan/utils.py
DasSkelett: I'm starting to really like type hints in Python, they are very useful.
Not only does my IDE complain immediately for wrong types, it also enables autocomplete if it isn't able to infer the type automatically.

PR #119, opened 2020-01-21
Adder Deployment Tweaks

techman83:
This tells travis to restart the service + sets the default region for it.


techman83:
Authentication is automatic via the Assume Role stuff, but region is not (though you can pull the current region from the metadata), because you might be in one region but need to talk to a different one. Being explicit about where you want to talk to is important as there can be implications relating to billing and what not if you're in one region and doing stuff in another.

PR #121, opened 2020-01-23
Fix Adder Path + Adjust Memory

techman83:
The Adder PR was merged not long before this and thus missed getting the changes.


techman83:
Merging to fix current spray of logs into the discord.

PR #126, opened 2020-01-28
Mirrorer Switchover

techman83:
This is a formalisation at this point, the service needed a couple of tweaks to get it to come up properly. This PR matches what's currently deployed in production.

### DIFF
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -354,11 +354,16 @@ def auto_freezer(netkan_remote, token, repo, user, days_limit, key):
     help='Collection to put mirrored mods in on Internet Archive',
     required=True,
 )
-def mirrorer(queue, timeout, ckan_meta, ia_access, ia_secret, ia_collection):
+@click.option(
+    '--key', envvar='SSH_KEY', required=True,
+    help='SSH key for accessing repositories',
+)
+def mirrorer(queue, timeout, ckan_meta, ia_access, ia_secret, ia_collection, key):
+    init_ssh(key, Path(Path.home(), '.ssh'))
     Mirrorer(
-        init_repo(ckan_meta, '/tmp/CKAN-meta'),
+        init_repo(ckan_meta, '/tmp/CKAN-meta'), queue,
         ia_access, ia_secret, ia_collection
-    ).process_queue(queue, timeout)
+    ).process_queue(timeout)
 
 
 @click.command()
--- a/netkan/netkan/mirrorer.py
+++ b/netkan/netkan/mirrorer.py
@@ -17,8 +17,10 @@ class Mirrorer:
     EPOCH_ID_REGEXP = re.compile('-[0-9]+-')
     EPOCH_TITLE_REGEXP = re.compile(' - [0-9]+:')
 
-    def __init__(self, ckan_meta_repo, ia_access, ia_secret, ia_collection):
+    def __init__(self, ckan_meta_repo, queue, ia_access, ia_secret, ia_collection):
+        sqs = boto3.resource('sqs')
         self.sqs_client = boto3.client('sqs')
+        self.queue = sqs.get_queue_by_name(QueueName=queue)
         self.ckan_meta_repo = ckan_meta_repo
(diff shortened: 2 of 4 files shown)


netkan/netkan/cli.py
```diff
@@ -354,11 +354,16 @@ def auto_freezer(netkan_remote, token, repo, user, days_limit, key):
...
+    help='SSH key for accessing repositories',
+)
+def mirrorer(queue, timeout, ckan_meta, ia_access, ia_secret, ia_collection, key):
+    init_ssh(key, Path(Path.home(), '.ssh'))
     Mirrorer(
-        init_repo(ckan_meta, '/tmp/CKAN-meta'),
+        init_repo(ckan_meta, '/tmp/CKAN-meta'), queue,
         ia_access, ia_secret, ia_collection
-    ).process_queue(queue, timeout)
+    ).process_queue(timeout)
```
HebaruSan: Can we put the queue param back where it was, since `purge_epochs` doesn't need it?

netkan/netkan/mirrorer.py
```diff
@@ -17,8 +17,10 @@ class Mirrorer:
     EPOCH_ID_REGEXP = re.compile('-[0-9]+-')
     EPOCH_TITLE_REGEXP = re.compile(' - [0-9]+:')
 
-    def __init__(self, ckan_meta_repo, ia_access, ia_secret, ia_collection):
+    def __init__(self, ckan_meta_repo, queue, ia_access, ia_secret, ia_collection):
```
HebaruSan: `cli.mirror_purge_epochs` doesn't pass queue.

netkan/netkan/cli.py
techman83: It's going to get refactored when I rebase the CLI. We can cross the epoch bridge when we look closer at that. Open an issue for now, as this is just a "get it running" pr.

netkan/netkan/mirrorer.py
techman83: Yep, I only changed what was needed to get it running. That's going to need changes for the CLI as well, so will be addressed when I rebase that work

HebaruSan: But this means `cli.mirror_purge_epochs` is broken. Shouldn't that PR be rebased on working code?

netkan/netkan/cli.py
```diff
@@ -354,11 +354,16 @@ def auto_freezer(netkan_remote, token, repo, user, days_limit, key):
     help='Collection to put mirrored mods in on Internet Archive',
     required=True,
 )
-def mirrorer(queue, timeout, ckan_meta, ia_access, ia_secret, ia_collection):
+@click.option(
+    '--key', envvar='SSH_KEY', required=True,
+    help='SSH key for accessing repositories',
+)
+def mirrorer(queue, timeout, ckan_meta, ia_access, ia_secret, ia_collection, key):
+    init_ssh(key, Path(Path.home(), '.ssh'))
```
HebaruSan: Why do we need the SSH key if we don't need to push?

netkan/netkan/mirrorer.py
techman83: I'm comfortable with it remaining broken. It's going to need refactoring to fit anyway. In truth it probably belonged as a code snippet in a proposal about handling epochs and the archive better. But I wasn't going to hold up this PR for something that will be a manually run utility.

This PR is to line up the repo with production, any changes beyond that don't really fit the scope.

netkan/netkan/cli.py
techman83: We're accessing the repo via SSH, so it needs a key. It's something we should consider though and would be pretty straight forward to do differently with the CLI refactor. No ssh key, provide a https endpoint, SSH key, provide an SSH endpoint.

techman83:
Added a commit to address resourcing to allow the scheduler to run.

netkan/netkan/mirrorer.py
HebaruSan: OK, will create an issue instead.

PR #129, opened 2020-01-28
Fix mirror filenames and epochs

### LINKED ISSUE #128: [Bug] Problems with queue param of Mirrorer
## Problems

#126 moved the `queue` param of `Mirrorer.process_queue` into the constructor.

https://github.com/KSP-CKAN/NetKAN-Infra/blob/3c64accf0eb10c14c3ef304908cbf123b87d7c5d/netkan/netkan/mirrorer.py#L20

- This is only needed by `process_queue`
- `cli.mirror_purge_epochs` doesn't pass it and so doesn't work

https://github.com/KSP-CKAN/NetKAN-Infra/blob/3c64accf0eb10c14c3ef304908cbf123b87d7c5d/netkan/netkan/cli.py#L394-L399

## Suggestions

Move the `queue` param back to `Mirrorer.process_queue`.

HebaruSan:
## Problems

- https://archive.org/download/EditorExtensionsRedux-3.4.1 is the new mirrorer's first archive and https://archive.org/download/NovaPunchRebalanced-Thor-0.1.7.1 is the legacy mirrorer's last. Comparing them, the new code is using the wrong filename; the hash and extension are missing:
  image
  image
- The new mirrorer currently drops epochs from names, titles, filenames, etc. This is wrong because we're trying to implement "like-for-like" until we can devise an effective migration strategy for the client to remove epochs. Luckily it has not yet archived a module with an epoch.
- The `mirror_purge_epochs` command is broken, because `Mirrorer`'s constructor requires a `queue` parameter that it cannot provide

## Causes

- We're passing `CkanMirror.mirror_item()`, which is identifier-hyphen-version, as the key, which [the documentation](https://archive.org/services/docs/api/internetarchive/internetarchive.html#internetarchive.Item.upload_file) explains thusly:
  image
  Meanwhile the `CkanMirror.mirror_filename()` is not used at all.
- The `with_epoch` params had a default of `False`
- Emergency changes copied from prod in #126 to get the core mirrorer working

## Changes

- Now we pass `CkanMirror.mirror_filename()` as the key instead
- Now the default for `with_epoch` is `True`
- Now `Mirrorer.__init__`'s `queue` param is moved back to `process_queue`

Fixes #128.

### DIFF
--- a/netkan/netkan/cli.py
+++ b/netkan/netkan/cli.py
@@ -361,9 +361,9 @@ def auto_freezer(netkan_remote, token, repo, user, days_limit, key):
 def mirrorer(queue, timeout, ckan_meta, ia_access, ia_secret, ia_collection, key):
     init_ssh(key, Path(Path.home(), '.ssh'))
     Mirrorer(
-        init_repo(ckan_meta, '/tmp/CKAN-meta'), queue,
+        init_repo(ckan_meta, '/tmp/CKAN-meta'),
         ia_access, ia_secret, ia_collection
-    ).process_queue(timeout)
+    ).process_queue(queue, timeout)
 
 
 @click.command()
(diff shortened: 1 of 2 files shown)


HebaruSan:
Confirming filename fixed on latest upload:

image

PR #130, opened 2020-01-28
Fix SpaceDock API format and docs

HebaruSan:
## Problem

While investigating KSP-SpaceDock/SpaceDock#232, I noticed some errors in the new (not yet used) SpaceDock web hooks:

- `/sd/inflate`'s documentation makes it look like it receives its parameters from the URL separated by `?` and `=` and `&`, but what I intended (and implemented in code) was for it to match the existing SpaceDock-Notify project (so it could be a drop-in replacement), which retrieves inputs from the form parameters of the POST request
- `/sd/add` extracts its inputs from the POST body as JSON, but I meant for it to use the same method of receiving parameters as `/sd/inflate`, since this would be simpler for both CKAN and SD maintenance

## Changes

- Now `/sd/inflate`'s documentation is updated to make it clearer that its input is from the POST params
- Now `/sd/add` is updated to get its inputs from POST params, and its documentation is updated to match. [`request.form` is an `ImmutableMultiDict`](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Request.form), so it should be OK to pass it to `batch_message` which will make a string out of it using `json.dumps`.

These changes will expedite efforts to migrate SpaceDock over to these new APIs.

### DIFF
--- a/netkan/netkan/webhooks/spacedock_add.py
+++ b/netkan/netkan/webhooks/spacedock_add.py
@@ -10,26 +10,22 @@ spacedock_add = Blueprint('spacedock_add', __name__)  # pylint: disable=invalid-
 
 # For mod creation hook on SpaceDock, creates pull requests
 # Handles: https://netkan.ksp-ckan.space/sd/add
-# POST json input:
-#   {
-#     "name":              "Mod Name Entered by the User on SpaceDock",
-#     "id":                "12345",
-#     "license":           "GPL-3.0",
-#     "username":          "ModAuthor1",
-#     "email":             "modauthor1@gmail.com",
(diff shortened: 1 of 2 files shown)


DasSkelett:
Just to be sure I understand the `/sd/add` change correctly, it's still meant to receive a JSON body, we just don't explicitly pull it as such from flask anymore, right? So not a functional change?

HebaruSan:
No, there **is** a functional change for `/sd/add` (which is why it's important that this API isn't in use yet). Digging into the details of HTTP POST requests, it looks like the parameters will still be in the body, but not as JSON:

https://stackoverflow.com/questions/14551194/how-are-parameters-sent-in-an-http-post-request

PR #132, opened 2020-01-29
Check SHA1 of mirrored files

### LINKED ISSUE #131: [Bug] Mirrorer only Checks Item, not files
Problem
-------
There is an implementation change between the legacy mirroring and the new mirroring:

https://github.com/KSP-CKAN/NetKAN-bot/blob/master/lib/App/KSP_CKAN/Tools/IA.pm#L268

The perl method checks to see if the SHA1s of the files match before considering it mirrored.
```perl
  if ( any { defined $_->{sha1} && uc($_->{sha1}) eq $ckan->download_sha1 } @{$data->{files}} ) {
    return 1;
  }

  return 0;
```

However the new service only checks to see if the item exists:
```python
    def mirrored(self, iarchive):
        results = iarchive.search_items(self.mirror_item())
        return True if results else False
```

Files can change and be the same version or a file upload may not complete and we could miss retrying later.

Recommendation
-------
Ensure we check all the files to see if we have a matching SHA1. Though I'm in favour of switching to a different algorithm if better ones are available via the IA metadata now.

HebaruSan:
## Problem

`CkanMirror.mirrored()` currently returns `True` if its `Item` exists on archive.org even if the file is not there. The legacy mirror hook checked for a file with a matching SHA1. This is blocking the re-upload of EditorExtensionsRedux-3.4.1.

## Changes

Now `CkanMirror.mirrored()` only returns `True` if it finds a file with a matching SHA1, like the old hook did.

Fixes #131.

### DIFF
--- a/netkan/netkan/mirrorer.py
+++ b/netkan/netkan/mirrorer.py
@@ -278,8 +278,13 @@ class CkanMirror(Ckan):
         )
 
     def mirrored(self, iarchive):
-        results = iarchive.search_items(self.mirror_item())
-        return True if results else False
+        item = iarchive.get_item(self.mirror_item())
+        if not item:
+            return False
+        if not item.exists:
+            return False
+        sha1 = self.download_hash['sha1'].lower()
+        return any(file['sha1'].lower() == sha1 for file in item.files if 'sha1' in file)
 
     def license_urls(self):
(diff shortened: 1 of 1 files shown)


HebaruSan:
Oops, it's `file['sha1']`. Still getting thrown by Python not being Javascript.

PR #135, opened 2020-01-30
Fix scheduler tests

DasSkelett:
## Problem
By chance I discovered that we are not running the scheduler test in `tests/scheduler.py`.
My IDE was reporting some syntax problems, missing arguments and wrong imports, and I wondered how the tests would succeed this way.

## Cause
This is because the name for the test class used in `tests/scheduler.py` is also used in `tests/metadata.py`. `Unittest` seems to have problems with that (understandably), and doesn't touch that class when encountering it the second time.

## Changes
I've renamed the class.
I also did the minimal changes to get the tests working again. IMHO they need a bigger redo, for me it looks like they are currently not that useful / missing out some stuff to test.
They are also testing a lot of other modules and are overlapping with other test files. But this is out of the nature of the Netkan package.

I moved one test that didn't touch the scheduler at all in `metadata.py`.
I thought about moving the two unit tests that basically just test `netkan.common.sqs_batch_entries()` into a new `tests/common.py`, but they will need the scheduler to load the netkans anyway, so I left them where they are now.

So yeah, these tests should be revisited again, but I'll leave that to my future self or somebody else for now, because I need to think of a plan how to reasonably structure these tests first.


techman83:
Yep, we definitely need more testing work. A lot of the common functions we should write into the individual services using them, then refactor into a base class that they can inherit from.

Thanks for the efforts!

Issue #133, opened 2020-01-30
[Bug] Upload causes mirrorer to crash


techman83:
Problem
-------
Upload succeeds, but crashes the process when returning.
```
[2020-01-30 01:11:13,750] [ERROR   ] Uncaught exception:
OSError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".local/bin/netkan", line 8, in <module>
    sys.exit(netkan())
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli.py", line 366, in mirrorer
    ).process_queue(queue, timeout)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/mirrorer.py", line 52, in process_queue
    if self.try_mirror(CkanMirror(path)):
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/mirrorer.py", line 80, in try_mirror
    return self.try_upload(ckan, tmp.file)
  File "/usr/local/lib/python3.7/tempfile.py", line 499, in __exit__
    result = self.file.__exit__(exc, value, tb)
OSError: [Errno 9] Bad file descriptor
OSError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".local/bin/netkan", line 8, in <module>
    sys.exit(netkan())
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
[...]

PR #136, opened 2020-01-30
Archive source code

### LINKED ISSUE #111: Should we be archiving source code as well?
## Motivation

I was trying to find the source code to whatever is in KerbalGalaxy.dll because [someone asked about it on the forum](https://forum.kerbalspaceprogram.com/index.php?/topic/184416-*).

The original download and source links are on dropbox, now broken. After checking other forum threads, [a SpaceDock page](https://spacedock.info/mod/872), and so on, I was able to confirm that there is indeed a KerbalGalaxy.dll included in this mod, but the source code appears to be totally unavailable.

Finally I thought to check [the archive.org collection](https://archive.org/details/kspckanmods). [KerbalGalaxyRevamped is indexed](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/KerbalGalaxyRevamped.netkan) in CKAN ([CC-NC-SA-4.0 license](https://creativecommons.org/licenses/by-nc-sa/4.0/)), so it was there as well, but no source code.

This gave me pause. Some licenses (notably GPL) require distribution of source code in addition to compiled binaries. Normally CKAN presumes that the source will be available *somewhere* out there when uploading redistributable mods to archive.org, but this KerbalGalaxy situation shows that this isn't always the case. If we indexed a mod with a GPL-style license, uploaded it to archive.org, and then the upstream source was subsequently withdrawn, would that violate the license?

## Suggestion

Should the bot upload source code to archive.org as well? This would ensure that it would remain available. We could limit this to licenses that require it.

HebaruSan:
## Motivation

`Mirrorer` uploads mods with redistributable licenses to archive.org. However, some redistributable licenses are conditional; with the GPL, the source code must be available. If an author indexes a GPL-licensed mod on CKAN and then deletes the original source code repo, then it's not clear whether it's OK for our mirror to remain on archive.org.

## Changes

Now we attempt to archive source code as well.

- A bunch of logic is moved from `Mirrorer` to `Ckan` and `CkanMirror` ([~~which reminds me I forgot to add tests, will do that in a bit~~](https://github.com/KSP-CKAN/NetKAN-Infra/pull/136/commits/0f2b0afa584b6738a660be3bc6088487cc33986c))
- New property `Ckan.source_download` gives the URL to download a module's source based on `resources.repository`
- Most of what was once `Mirrorer.try_upload` is now farmed out to helper functions in `CkanMirror`, leaving behind only the core of the uploading logic, which is moved into `Mirrorer.try_mirror`
- `Mirrorer.try_mirror` now also has a new block that retrieves the source code and uploads it to the same item as an additional file
- If we download a module to a temp file, we now copy it to the cache and upload from the cache rather than uploading the temp file. This may or may not help with #133.

Fixes #111.

### DIFF
--- a/netkan/netkan/mirrorer.py
+++ b/netkan/netkan/mirrorer.py
@@ -50,7 +50,7 @@ class Mirrorer:
                 to_delete = []
                 for msg in messages:
                     path = Path(self.ckan_meta_repo.working_dir, msg.body)
-                    if self.try_mirror(CkanMirror(path, self.ia_collection)):
+                    if self.try_mirror(CkanMirror(self.ia_collection, path)):
                         # Successfully handled -> OK to delete
                         to_delete.append(self.deletion_msg(msg))
                 queue.delete_messages(Entries=to_delete)
@@ -247,8 +247,8 @@ class CkanMirror(Ckan):
 
     EPOCH_VERSION_REGEXP = re.compile('^[0-9]+:')
 
-    def __init__(self, filename, collection):
-        Ckan.__init__(self, filename)
+    def __init__(self, collection, filename=None, contents=None):
+        Ckan.__init__(self, filename, contents)
         self.collection = collection
 
     @property
--- a/netkan/netkan/mirrorer.py
+++ b/netkan/netkan/mirrorer.py
@@ -1,8 +1,10 @@
+import os
 import re
 import tempfile
 import urllib.request
 import hashlib
 import logging
+import shutil
 from pathlib import Path
 import requests
 import boto3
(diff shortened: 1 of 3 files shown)


netkan/netkan/mirrorer.py
```diff
@@ -313,12 +295,62 @@ def mirror_filename(self, with_epoch=True):
...
+            return cached_file.open(mode='rb')
+        # Download the file as needed
+        with tempfile.NamedTemporaryFile() as tmp:
+            logging.info('Downloading %s', self.download)
+            urllib.request.urlretrieve(self.download, tmp.name)
+            # Copy to cache so the temp file can be deleted
+            shutil.copyfile(tmp.name, self.CACHE_PATH.joinpath(self.cache_filename))
+            logging.info('Downloaded %s to %s', self.mirror_item(), self.cache_filename)
+        return self.cache_find_file.open(mode='rb')
+
```
techman83: I don't think we want to cache mirrored files, it's unlikely they'll be used by the inflator. But this should fix the current issue, so lets worry about that later.

PR #146, opened 2020-02-09
Fix branch check

HebaruSan:
## Problem

Mirroring stopped 4-5 days ago.

https://archive.org/details/kspckanmods?sort=-publicdate

## Cause

We used the string `ref/heads/master` in #144 to represent the master branch. Git uses `refs/heads/master`, so right now we consider all web hook requests to be on the wrong branch.

Compare the same check for SpaceDock:

https://github.com/KSP-SpaceDock/SpaceDock/blob/ccb56f7d876ebf40057669fa60720d29e62f5d78/KerbalStuff/app.py#L152-L157

## Changes

Now it says `refs/heads/master` instead.

### DIFF
--- a/netkan/netkan/webhooks/github_mirror.py
+++ b/netkan/netkan/webhooks/github_mirror.py
@@ -16,10 +16,11 @@ github_mirror = Blueprint('github_mirror', __name__)  # pylint: disable=invalid-
 @signature_required
 def mirror_hook():
     raw = request.get_json(silent=True)
-    branch = raw.get('ref')
-    if branch != 'ref/heads/master':
+    ref = raw.get('ref')
+    expected_ref = current_app.config['ckanmeta_repo'].heads.master.path
+    if ref != expected_ref:
         current_app.logger.info(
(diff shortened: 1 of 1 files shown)


DasSkelett:
For reference, the webhooks documentation: https://developer.github.com/v3/activity/events/types/#events-api-payload-34
> ## PushEvent
> ### Events API payload
> Key | Type | Description
> -- | -- | --
> ref | string | The full git ref that was pushed. Example: refs/heads/master.

HebaruSan:
I'm investigating a possible way to generate this string via the GitPython API instead of hard coding it. I'll reply once I've determined whether it's viable.

HebaruSan:
```python3
#!/usr/bin/env python3

from git import Repo

r = Repo('.')
print(r.heads.master.path)
```

```bash
$ print-master-ref.py 
refs/heads/master
```

techman83:
Go ahead and merge it if it looks good, my cursory glance doesn't show anything obvious wrong.

PR #147, opened 2020-02-11
Enable `init` on Indexer

techman83:
ECS has support for linux parameters, one of them being the `init` flag. Docker included tini as an init option, that will take care of the usual tasks an init system would do, in this case reaping zombie processes.

Long term using GitPython may not be sustainable as we lose visibility to problems as it uses the git binary directly. But this should avoid the indexer crashing every other day due to 1000s of defunct ssh sessions.

*Note:* This was a stack update and has already been manually applied and is working.

```
~$ ps waux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
netkan       1  0.1  0.0    944     8 ?        Ss   02:05   0:00 /dev/init -- .local/bin/netkan indexer
netkan       7  1.0  5.6  79240 55488 ?        S    02:05   0:01 /usr/local/bin/python .local/bin/netkan indexer
netkan      13  7.5  0.3   5756  3492 pts/0    Ss   02:06   0:00 bash
netkan      19  0.0  0.2   9396  2900 pts/0    R+   02:06   0:00 ps waux
```

### DIFF
--- a/prod-stack.py
+++ b/prod-stack.py
@@ -7,7 +7,7 @@ from troposphere.dynamodb import Table, KeySchema, AttributeDefinition, \
     ProvisionedThroughput
 from troposphere.ecs import Cluster, TaskDefinition, ContainerDefinition, \
     Service, Secret, Environment, DeploymentConfiguration, Volume, \
-    Host, MountPoint, PortMapping, ContainerDependency
+    Host, MountPoint, PortMapping, ContainerDependency, LinuxParameters
 from troposphere.ec2 import Instance, CreditSpecification, Tag, \
     BlockDeviceMapping, EBSBlockDevice
 from troposphere.cloudformation import Init, InitFile, InitFiles, \
@@ -605,6 +605,7 @@ services = [
         'volumes': [
             ('ckan_cache', '/home/netkan/ckan_cache')
         ],
+        'linux_parameters': LinuxParameters(InitProcessEnabled=True),
     },
     {
         'name': 'Scheduler',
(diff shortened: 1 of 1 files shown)


HebaruSan:
I think we should apply this to all containers.

techman83:
Unless there is a good reason to, I don't think we should. The indexer is the only process to exhibit this issue so far.

HebaruSan:
The reason would be that most of our containers use GitPython to access `ssh://` based remotes which will eventually run into this problem.

What's the downside to fixing all containers?

techman83:
None of the other containers have exhibited this problem. I'm not fully across the consequences of using this init system and as far as I'm concerned this is a work around, not a fix. I don't think we should apply it unilaterally, especially not in an off the cuff "lets see if this fixes it" PR.

HebaruSan:
It's a fix, but to a core problem with the concept of containers (missing critical parts of UNIX), as opposed to a problem with python or git or ssh. Your link explained it:

- https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/

The reluctance of the git devs on the thread @DasSkelett found to fix the issue illustrates how core this functionality is and therefore how serious it is to not have a zombie reaper process:

- https://lore.kernel.org/git/20190722212250.44011-1-thibault.jamet@gmail.com/

This is affecting git and ssh, *now*. Next time it can affect some other library, or some GNU utility, etc. There's no way to predict it in advance, because a core system function is missing.

techman83:
I don't think any of that applies to this PR. If we think there is a fundamental issue with the way containers run, it belongs in an Issue for more thorough discussion and planning. In my experience this is a pretty unusual problem and relates directly to how we're using git, which is covered in #148. In fact by enabling it may hide problems that we should be aware of, hence my apprehension.

techman83:
That's a full run without any defunct processes hanging around. I think this is an acceptable work around for now, I think #148 should be the long term goal, but this should hopefully stop the box thrashing the disk until it can no longer create new processes every couple of days.

PR #149, opened 2020-02-11
Don't mount cache for Indexer

HebaruSan:
Currently the Indexer container has access to the shared download cache, but it doesn't need or use it. This probably doesn't cause any problems (other than making the infra slightly more confusing).

This PR removes the cache volume from the Indexer definition.
(Really just submitting this to get it out of my persistent git stash that I carry from one branch to the next.)


techman83:
Haha, yeah. Not even sure why it ended up there in the first place :slightly_smiling_face:

PR #152, opened 2020-02-15
Fix AttributeError in Mirrorer for ckans without download_content_type

DasSkelett:
## Problem
https://github.com/KSP-CKAN/CKAN-meta/pull/1819 merged some changes to ckan files that didn't have `download_content_type` specified.
We got these exceptions thrown by the mirrorer:
```
Uncaught exception:
    sys.exit(netkan())
    rv = self.invoke(ctx)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/decorators.py", line 64, in new_func
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/services.py", line 72, in mirrorer
    ).process_queue(common.queue, common.timeout)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/mirrorer.py", line 53, in process_queue
    if self.try_mirror(CkanMirror(self.ia_collection, path)):
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/mirrorer.py", line 61, in try_mirror
    if not ckan.can_mirror:
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/metadata.py", line 110, in __getattr__
    raise AttributeError
AttributeError
```
Although the stack trace makes it look like Python tries call `__getattr__` when looking for `ckan.can_mirror`, I suspect it actually _finds_ `can_mirror` and throws the error inside `can_mirror` when trying to access `self.download_content_type`, which then calls `__getattr__` of the `Ckan` class, which raises an `AttributeError` if the source JSON doesn't have that property (which, in this case, is true for `download_content_type`).

## Changes
Now we use `getattr(self, 'download_content_type', '')` in `can_mirror`.
This will return `False`, so the files don't get mirrored:
```
>>> '' in {
...         'application/x-gzip':           'gz',
...         'application/x-tar':            'tar',
...         'application/x-compressed-tar': 'tar.gz',
...         'application/zip':              'zip',
...     }
False
```

### DIFF
--- a/netkan/netkan/mirrorer.py
+++ b/netkan/netkan/mirrorer.py
@@ -255,7 +255,7 @@ class CkanMirror(Ckan):
     def can_mirror(self):
         return (
             self.kind == 'package'
-            and self.download_content_type in Ckan.MIME_TO_EXTENSION
+            and getattr(self, 'download_content_type', '') in Ckan.MIME_TO_EXTENSION
             and self.redistributable
         )
 


techman83:
Looks like a reasonable solution to me. For bug fixes + operational stuff, go ahead and do what you think is best. The CI/CD is specifically in place to enable that :slightly_smiling_face:

PR #154, opened 2020-02-18
Generate random unique batch ids

HebaruSan:
## Problem

The Scheduler is only scheduling A–K:

```
Uncaught exception:
    sys.exit(netkan())
    rv = self.invoke(ctx)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/decorators.py", line 64, in new_func
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/services.py", line 61, in scheduler
    sched.schedule_all_netkans()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/scheduler.py", line 48, in schedule_all_netkans
    self.client.send_message_batch(**self.sqs_batch_attrs(batch))
  File "/home/netkan/.local/lib/python3.7/site-packages/botocore/client.py", line 316, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/botocore/client.py", line 626, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.BatchEntryIdsNotDistinct: An error occurred (AWS.SimpleQueueService.BatchEntryIdsNotDistinct) when calling the SendMessageBatch operation: Id Kopernicus repeated.
```

## Cause

After #153, `NetKAN/NetKAN/Backports/Kopernicus.netkan` and `NetKAN/NetKAN/Kopernicus.netkan` ended up in the same batch because we sort by the file stem. Since they have the same identifier, the identifier isn't a good enough `Id` anymore.

## Changes

Now we use random hex strings, just like in #81.

### DIFF
--- a/netkan/netkan/metadata.py
+++ b/netkan/netkan/metadata.py
@@ -75,11 +75,12 @@ class Netkan:
         return attribs
 
     def sqs_message(self, ckan_group=None):
+        id = uuid.uuid4().hex
         return {
-            'Id': self.identifier[0:80],
+            'Id': id,
             'MessageBody': self.contents,
             'MessageGroupId': '1',
-            'MessageDeduplicationId': uuid.uuid4().hex,
+            'MessageDeduplicationId': id,
             'MessageAttributes': self.sqs_message_attribs(ckan_group),
         }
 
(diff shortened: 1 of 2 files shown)


techman83:
It's not really a big deal if we happen generate the same item twice. But we may want to revisit the dedupe stuff at some point. We use FIFOs more to avoid caring about reliability of messages being put on the queue and being removed, rather than say a scenario where we have multiple things processing and producing messages that could clash.

Issue #148, opened 2020-02-11
[Feature] PyGit2


techman83:
Problem
-------
GitPython shells out to git, which has benefits and consequences. Benefit being that it closely matches what git can already do and you can run git commands if you need to. However you lose some visibility over the process and it's leaky, as noted in #142 

Suggestions
-----------
PyGit2 has matured rapidly since the service was built and uses the libgit2 C bindings directly. We would have a lot more visibility over what's going on and wouldn't need to mess around with our environment as much (you can pass keys/user details etc straight into pygit2).

HebaruSan:
The repo says https://www.pygit2.org/ is their documentation, but it's pretty horrible. Is there an actual intro/tutorial somewhere, with simple yet complete examples?

HebaruSan:
> (you can pass keys/user details etc straight into pygit2)

Unfortunately it looks like SSH connections still require creation of files:

https://www.pygit2.org/remotes.html#pygit2.Keypair

![image](https://user-images.githubusercontent.com/1559108/74593158-14eb5800-4fee-11ea-903e-46cca4e983ce.png)

techman83:
Yeah, I think stuff like that is why I stuck with GitPython. I didn't want to get bogged down yak shaving stuff like this. Now everything is kinda done, refactoring our git interactions is a little less daunting.

We might want to move our git interactions into a mixin class or something, with some tests. Then we can play around with pygit2 a little easier.

HebaruSan:
In the meantime, should we look into periodically restarting the various containers? The latest pass is taking **9 hours** to get from 'A' to 'O', as compared to the 10-20 minutes for the entire pass that we had a short while ago.

techman83:
![2020-06-15_10-30-55](https://user-images.githubusercontent.com/3467784/84612525-787d8780-aef3-11ea-91bb-81682820d40b.png)

A refactor was done in #178, along with the work fixing #142 this is less of a priority. There is still more refactor work to do, but I think we can close this as an issue.

PR #156, opened 2020-04-10
Regularly restart GitPython-based services

HebaruSan:
## Problems

The bot has been skipping a lot of runs for months now based on the checks from #112 and #115:

image

And the current pass is taking an extremely long time to complete, over 10 hours to get from 'A' to 'P' (where it used to be 10-20 minutes for 'A' to 'Z'):

image

image

This has meant a very long delay in the post-merge handling of KSP-CKAN/NetKAN#7831 (and who knows what else).

## Cause

We still don't know exactly why this happens, but based on graphs that @techman83 has shared previously, it seems to be something that starts out mild and then gets gradually worse the longer the containers run. May be due to some aspect of GitPython, but we don't quite have proof of that, and rewriting to use another tool is a big project (see #148).

## Changes

There's already a RestartWebhooks service to handle changes in certs once a week.

Now some new duplicates of that service are created to automatically restart the Indexer, Adder, and Mirrorer every 3 days. This should reset whatever the problem is with the Python-based services and hopefully give us better service availability.

### DIFF
--- a/prod-stack.py
+++ b/prod-stack.py
@@ -346,7 +346,8 @@ netkan_ecs_role = t.add_resource(Role(
 scheduler_resources = []
 for task in [
         'Scheduler', 'SchedulerWebhooksPass', 'CertBot', 'StatusDumper',
-        'DownloadCounter', 'TicketCloser', 'AutoFreezer', 'RestartWebhooks']:
+        'DownloadCounter', 'TicketCloser', 'AutoFreezer', 'RestartWebhooks',
+        'RestartAdder', 'RestartIndexer', 'RestartMirrorer']:
     scheduler_resources.append(Sub(
         'arn:aws:ecs:*:${AWS::AccountId}:task-definition/NetKANBot${Task}:*',
         Task=task
--- a/prod-stack.py
+++ b/prod-stack.py
@@ -714,6 +714,42 @@ services = [
         'schedule': 'cron(30 0 ? * MON *)',
     },
     {
+        'name': 'RestartIndexer',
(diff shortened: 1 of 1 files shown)


techman83:
The only thing is, we'd have to to restart these much more frequently than every 3 days. But we can implement it and now that things have settled at work, I should be able to get back to looking at pygit.

I'll check this out in the next couple of days if I don't think of something else that might give us some breathing space. I can tell you though that it's the Indexer where this problem lay and believe it is to do with all the branching as the other containers do lots of git work without issue and that's one of the key differences. Though the reflog may be also an issue, but I can't see how it could be.

HebaruSan:
KSP-CKAN/CKAN#3021 and KSP-CKAN/CKAN#3031 seem to have mostly solved the skipped runs problem by reducing the amount of staging, which @techman83 reports is important because git branch operations are costly and performed regardless of whether the files have changed. I think we can set this aside for now. Even if the problem returns, we know we can try to attack it by reducing usage of staging.

techman83:
I don't know that we should reduce our use of staging, but maybe get a bit smarter about creating branches. Though staging implies manual intervention, so being smart about when we stage is likely just as useful for less technical reasons.

HebaruSan:
> being smart about when we stage is likely just as useful for less technical reasons

Yup, my motivation for those PRs was not having to deal with unnecessary decisions. It was sheer dumb luck that it happened to also help with other problems. But now we know that if any future change increases the use of staging, there's a trade-off to be considered.

PR #159, opened 2020-04-21
Fix redeploy webhooks job

### LINKED ISSUE #158: [Bug] Scheduled Restarts of Webhooks not Restarting
Problem
-------
This is supposed to get restarted weekly (shortly after the cert renewal is checked)
```
50c539a33a62        kspckan/webhooks-proxy           "nginx -g 'daemon of…"   2 months ago        Up 2 months             0.0.0.0:80->80/tcp, 0.0.0.0:443->443/tcp
29f80c1275ef        kspckan/netkan                   ".local/bin/gunicorn…"   2 months ago        Up 2 months
```
Certs do get renewed, but 2 months is about when we last did a prod deployment.

HebaruSan:
## Background

The service redeploy command uses a `--service-name` argument, which is generally passed the name of the service without a `Service` suffix:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/a374d2187f38285e9f6306bc8cb9a99d97794851/bin/push_deploy_containers.sh#L18-L19

https://github.com/KSP-CKAN/NetKAN-Infra/blob/a374d2187f38285e9f6306bc8cb9a99d97794851/bin/push_deploy_containers.sh#L24-L25

https://github.com/KSP-CKAN/NetKAN-Infra/blob/a374d2187f38285e9f6306bc8cb9a99d97794851/bin/push_deploy_containers.sh#L30-L31

https://github.com/KSP-CKAN/NetKAN-Infra/blob/a374d2187f38285e9f6306bc8cb9a99d97794851/bin/push_deploy_containers.sh#L36-L37

https://github.com/KSP-CKAN/CKAN/blob/16994590ee0318d6bb93c90455e1dead093a02cc/build.cake#L88-L90

## Problem

The job to restart the webhooks once per week is not working, see #158.

## Cause

The prod stack setup file has a list of schedulable tasks, which does not include `RestartWebhooks`:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/a374d2187f38285e9f6306bc8cb9a99d97794851/prod-stack.py#L347-L349

As a point of minor inconsistency, the command for `RestartWebhooks` includes a `Service` suffix:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/a374d2187f38285e9f6306bc8cb9a99d97794851/prod-stack.py#L707-L709

## Changes

- Now `RestartWebhooks` is added to the list of schedulable tasks
- Now the scheduled task redeploys `Webhooks` instead of `WebhooksService`

Fixes #158.

### DIFF
--- a/prod-stack.py
+++ b/prod-stack.py
@@ -706,7 +706,7 @@ services = [
         'command': [
             'redeploy-service',
             '--cluster', 'NetKANCluster',
-            '--service-name', 'WebhooksService',
+            '--service-name', 'Webhooks',
         ],
         'env': [
             ('AWS_DEFAULT_REGION', Sub('${AWS::Region}')),
--- a/prod-stack.py
+++ b/prod-stack.py
@@ -346,7 +346,7 @@ netkan_ecs_role = t.add_resource(Role(
 scheduler_resources = []
 for task in [
         'Scheduler', 'SchedulerWebhooksPass', 'CertBot', 'StatusDumper',
-        'DownloadCounter', 'TicketCloser', 'AutoFreezer']:
+        'DownloadCounter', 'TicketCloser', 'AutoFreezer', 'RestartWebhooks']:
     scheduler_resources.append(Sub(
         'arn:aws:ecs:*:${AWS::AccountId}:task-definition/NetKANBot${Task}:*',
         Task=task


techman83:
The service name is generated automatically and contains 'WebhooksService'
```
NetKANBot-WebhooksService-JKP9JCTBJ9L8
```
Looking closer at the schedule, it's running but failing. It's very likely that the task itself doesn't have permission to restart the webhooks.

Looking at the schedule, it doesn't have a task definition role, so that's probably what needs fixing.
```
 Task Definition
NetKANBotSchedulerWebhooksPass:9
Launch type
EC2
ECS target
NetKANCluster
Target role
NetKANBot-NetKANProdSchedulerRole-VZVYE9SA340K
Number of tasks
1
Task definition role
--
Task role override
--
```

HebaruSan:
Why are the post-commit GitHub hooks able to successfully redeploy? Do they have permissions somehow?

HebaruSan:
Added a new commit, is that what you had in mind?

techman83:
Quite likely. I'll print out a template and see what it comes out like. Thanks for chipping away at this.

techman83:
Yep. I think you're on the money here. The task itself works, but the schedule wasn't firing. And well it didn't have permission to run the task, so it indeed looks like I missed allowing it! Good catch, I'll merge and deploy.

PR #161, opened 2020-05-04
Fix dict accesses in spacedock_adder

DasSkelett:
# Problem
The usual dict access problem :)
Dicts need accessing by `dict.get('something')`, not `dict.something`

# Changes
Replaced every dict access in `spacedock_adder.py` with a `.get()` access.

### DIFF
--- a/netkan/netkan/spacedock_adder.py
+++ b/netkan/netkan/spacedock_adder.py
@@ -72,9 +72,9 @@ class SpaceDockAdder:
 
         # Commit
         self.netkan_repo.index.commit(
-            f'Add {info.name} from {info.site_name}'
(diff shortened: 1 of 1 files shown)


HebaruSan:
I wish I had never learned Javascript.

DasSkelett:
Huh, don't understand the error message, I think it's unrelated:
`  error: could not create 'build': Permission denied`

HebaruSan:
Yeah, I don't think these changes could break the Docker build like that. But if there's a problem with it now, that might also cause problems after merge. Probably best to have @techman83 take a look.

HebaruSan:
On second thought, let's just merge in case it does help. If not we can investigate then.

PR #168, opened 2020-05-13
Notify on SpaceDock deletion

HebaruSan:
## Motivation

While working on KSP-SpaceDock/SpaceDock#257, I finally spotted the other `event_type`s that SpaceDock can send via the webhooks: `version-update` and `delete`.

`version-update` can be treated the same as a normal inflation request, but `delete` probably shouldn't be.

## Changes

- Now the `event_type` parameter is fully documented.
- Now if a `delete` request comes in, we log an error so we can see it on Discord.
  - I'm not 100% happy with logging an error for this, please suggest something better.

### DIFF
--- a/netkan/netkan/webhooks/spacedock_inflate.py
+++ b/netkan/netkan/webhooks/spacedock_inflate.py
@@ -12,7 +12,9 @@ spacedock_inflate = Blueprint('spacedock_inflate', __name__)  # pylint: disable=
 # Handles: https://netkan.ksp-ckan.space/sd/inflate
 # POST form parameters:
 #     mod_id:     The mod's ID number on SpaceDock
-#     event_type: update
+#     event_type: update         - New version of mod was uploaded
+#                 version-update - Default version changed
+#                 delete         - Mod was deleted from SpaceDock
 @spacedock_inflate.route('/inflate', methods=['POST'])
 def inflate_hook():
     # Make sure our NetKAN and CKAN-meta repos are up to date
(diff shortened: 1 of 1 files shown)


DasSkelett:
> * Now if a `delete` request comes in, we log an error so we can see it on Discord.
>     * I'm not 100% happy with logging an error for this, please suggest something better.

I guess you are thinking of reducing it to `logging.warning`, since it is not a real error?
If we change the level of the `DiscordLogHandler` to `logging.WARNING`, it should also send warning messages to Discord.
The result would be basically the same, but we could use both logging levels to keep the code "clean".
And if we ever want to, we can create another log handler for warning messages in the future, which send them to a different channel, differently formatted or whatever.

HebaruSan:
> I guess you are thinking of reducing it to `logging.warning`, since it is not a real error?

I was thinking that going through the logger is a cheat, but it is such an easy solution here. The proper approach if we want to send messages to Discord would be to set up some kind of explicit service for that separate from the logger.

DasSkelett:
> I was thinking that going through the logger is a cheat, but it is such an easy solution here. The proper approach if we want to send messages to Discord would be to set up some kind of explicit service for that separate from the logger.

Ah okay. Well, we could extract the webhook part from `DiscordLogHandler` in to a new `send_message_to_discord()` and call that from inside the log handler as well as externally, for example in this occasion.

Not entirely a different service, but a different class/method at least :)

But I'm also okay with cheating with the logging mechanism.

techman83:
Agreed, we can always refactor and it's a nice to have, rather than a core function.

PR #172, opened 2020-05-22
Encapsulate repo structure assumptions

HebaruSan:
## Motivation

The NetKAN and CKAN-meta repos have very specific structures.

- All netkans are in the NetKAN subfolder
- All unfrozen netkan filenames are of the form "identifier.netkan"
- Frozen netkan filenames are of the form "identifier.frozen"
- All ckans are in a folder named after their identifier
- And so on

Currently these assumptions are simply embedded in code wherever they are needed, which is not great for maintenance.

## Changes

Now a new `repos.py` file contains a `NetkanRepo` class and a `CkanMetaRepo` class, which provide public functions that centralize the above assumptions in one place, and existing code is updated to use them. If another class needs to access files in NetKAN or CKAN-meta, it can use these functions to know it's doing it right. If we ever need to change the structure of these repos, we'll be able to do it without auditing the entire codebase.


HebaruSan:
#rebuild

HebaruSan:
I made a third attempt at what we discussed in Discord, making the `NetkanRepo` and `CkanMetaRepo` objects own their respective `Repo` objects. I think that's done now.

netkan/netkan/cli/common.py
```diff
@@ -91,23 +90,31 @@ def ssh_key(self, value):
         init_ssh(value, Path(Path.home(), '.ssh'))
         self._ssh_key = value
 
+    @property
+    def ckanmeta_repo(self):
+        if not self._ckanmeta_repo:
+            self._ckanmeta_repo = CkanMetaRepo(
+                init_repo(self._ckanmeta_remote, '/tmp/CKAN-meta', self.deep_clone))
```
techman83, line 97: We could move the Getter/Setter stuff into XKanaMetaRepo, which would clean up the need for us to repeat this in the CLI multiple times.

netkan/netkan/repos.py
```diff
@@ -0,0 +1,62 @@
+import logging
+from pathlib import Path
+from git import Repo
+
+from .metadata import Netkan, Ckan, CkanGroup
+
+
```
techman83, line 7: From the comment earlier, we could have a MetaRepo class, that both these classes inherit from.

We can then use a setter/getter for cloning the repo. Sorta like this:
```python
class MetaRepo:

    def __init__(self, git_repo):
        self._git_repo = None
    
    @property
    def git_repo(self):
        if not self._git_repo:
            self._git_repo = init_repo(self._netkan_remote, '/tmp/NetKAN', self.deep_clone)
        return self._git_repo


class NetkanRepo(MetaRepo):

    """
    Encapsulates all assumptions we make about the structure of NetKAN
    """
    
    NETKAN_DIR = 'NetKAN'
    UNFROZEN_SUFFIX = 'netkan'
    FROZEN_SUFFIX = 'frozen'
    NETKAN_GLOB = f'**/*.{UNFROZEN_SUFFIX}'

    @property
    def nk_dir(self):
        return Path(self.git_repo.working_dir, self.NETKAN_DIR)


class CkanMetaRepo(MetaRepo):

    """
    Encapsulates all assumptions we make about the structure of CKAN-meta
    """
    
    CKANMETA_GLOB = '**/*.ckan'

    @property
    def ckm_dir(self):
        return Path(self.git_repo.working_dir)
```
Then we can abstract our git interactions, making it easier to potentially replace GitPython

techman83:
Comments added, I don't think they need addressing in this PR. Merge and we can continue to refactor smaller chunks as we go.

netkan/netkan/repos.py
HebaruSan: The wrinkle with having the `*kanRepo` classes call `init_repo` (which is what I did at first) is making that work with the Indexer tests:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/4b5e1d65f66e0209025fef0e5ca1e7c087d9427d/netkan/tests/indexer.py#L40-L50

The repo is set up by the test setup itself using `git` operations, and the `*kanRepo` class needs to be able to receive that and work with it. Hence having its `__init__` take a `Repo` as a parameter is convenient and powerful.

netkan/netkan/cli/common.py
HebaruSan: Well, we still need to _create_ the `*KanRepo` objects somewhere. And we're not repeating, here; the remote and path parameters are different in each call.

netkan/netkan/repos.py
HebaruSan: A shared parent class `__init__` won't set `self.nk_dir` or `self.ckm_dir`, which the classes need to operate properly. So they would each need `__init__`s anyway, which would call the parent `__init__` and then set those vars. Which would be the same number of lines of code, plus the extra lines to define the parent class. I don't think there's enough common logic to make this worth it.

HebaruSan:
I'm guessing we'll get breakage and exceptions sent to Discord, but I'll try to react to those in a timely fashion.

netkan/netkan/repos.py
techman83: > The repo is set up by the test setup itself using git operations, and the *kanRepo class needs to be able to receive that and work with it. Hence having its __init__ take a Repo as a parameter is convenient and powerful.

We could still achieve this. Set repo as a kwarg, default to `None`. During `__init__` do a check and set it if is populated. The property will just return it if it's populated.

> A shared parent class __init__ won't set self.nk_dir or self.ckm_dir, which the classes need to operate properly. So they would each need __init__s anyway, which would call the parent __init__ and then set those vars. Which would be the same number of lines of code, plus the extra lines to define the parent class. I don't think there's enough common logic to make this worth it.

`self.nk_dir` + `self.ckm_dir` are converted to properties in my example. I don't see why they need to be initialised during class instantiation. The only reason you wouldn't want to do this is performance and you can do something similar to my repo example.

Things like branching, push, pull, commit etc are all common to both classes. A lot of that logic is buried in individual classes and could serve us better to centralise.

PR #173, opened 2020-05-23
Type hints and checking in CI

HebaruSan:
## Motivation

Python supports optional type hints, similar to declaring types in a compiled language. The type checker `mypy` uses these type hints to find issues in code, similar to a compiler.

- https://www.python.org/dev/peps/pep-0484/
- http://mypy-lang.org/

## Changes

- `mypy` installed by `setup.py`
- Run `mypy` when Docker builds the container
- Ignore the cache for git
- Add type hints for everything
- Manual concatenation of HTML strings for description sent to archive.org is replaced by a small Jinja2 template
- The Mirrorer now gets the actual default branch for GitHub repos instead of assuming `master`, using the API and token
- `CkanGroup` is now merged into `CkanMetaRepo` to avoid a circular import
- GitHub notifications about `.frozen` files in the NetKAN repo will now set `ModStatus.frozen = True` (previously we had to wait for the AutoFreezer to run)

### DIFF
--- a/netkan/Dockerfile
+++ b/netkan/Dockerfile
@@ -2,10 +2,11 @@ FROM python:3.7 as base
 RUN useradd -ms /bin/bash netkan
 ADD . /netkan
 WORKDIR /netkan
+RUN pip install mypy
 RUN chown -R netkan:netkan /netkan
 USER netkan
 RUN pip install --user . --no-warn-script-location
-RUN /home/netkan/.local/bin/mypy .
+RUN mypy .
 RUN python -m unittest -v
 RUN /home/netkan/.local/bin/netkan --help
 
@@ -23,9 +24,11 @@ CMD ["--help"]
 FROM production as dev
 USER root
 ADD . /netkan
+RUN chown -R netkan:netkan /netkan
 ADD run_dev.sh /usr/local/bin/
 USER netkan
 RUN pip install --user /netkan/.[development]
+RUN pip install --user /netkan/.[test]
 ENTRYPOINT ["/usr/local/bin/run_dev.sh"]
 
 FROM production
--- a/netkan/Dockerfile
+++ b/netkan/Dockerfile
@@ -5,6 +5,7 @@ WORKDIR /netkan
 RUN chown -R netkan:netkan /netkan
 USER netkan
 RUN pip install --user . --no-warn-script-location
+RUN /home/netkan/.local/bin/mypy .
 RUN python -m unittest -v
 RUN /home/netkan/.local/bin/netkan --help
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.egg-info
 /pr_tester
 __pycache__
+.mypy_cache/
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,6 +8,7 @@
     "--rcfile",
     "/home/netkan/workspace/.pylintrc"
   ],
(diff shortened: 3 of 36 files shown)


HebaruSan:
Definitely one of the suspected failure points. `mypy` as production dependency doesn't seem right, though.

```
/bin/sh: 1: mypy: not found
The command '/bin/sh -c mypy --ignore-missing-imports .' returned a non-zero code: 127
```

HebaruSan:
Hmm.

```
Collecting mypy
  Downloading mypy-0.770-cp37-cp37m-manylinux1_x86_64.whl (21.7 MB)
...
Collecting mypy-extensions<0.5.0,>=0.4.3
  Downloading mypy_extensions-0.4.3-py2.py3-none-any.whl (4.5 kB)
...
Installing collected packages: [...] mypy-extensions, typing-extensions, typed-ast, mypy, netkan-indexer
Successfully installed [...] mypy-0.770 mypy-extensions-0.4.3 [...]
...
/bin/sh: 1: mypy: not found
The command '/bin/sh -c mypy --ignore-missing-imports .' returned a non-zero code: 127
```

:snake:

HebaruSan:
:tada:

```
Step 8/20 : RUN /home/netkan/.local/bin/mypy --ignore-missing-imports .
 ---> Running in a6a033050b67
Success: no issues found in 35 source files
```

But we probably don't want it sticking around the container after build...

netkan/Dockerfile
```diff
@@ -5,6 +5,7 @@ WORKDIR /netkan
 RUN chown -R netkan:netkan /netkan
 USER netkan
 RUN pip install --user . --no-warn-script-location
+RUN /home/netkan/.local/bin/mypy .
 RUN python -m unittest -v
```
techman83, line 9: I'm assuming mypy is only used in tests? We could add a test section in our extra requires then pip install them globally in this stage. 

Like this
```
RUN pip install /netkan/.[test]
RUN /home/netkan/.local/bin/mypy .
￼RUN python -m unittest -v
```
Then these deps won't end up in our production container. The Dev stage will need them installed as well.

HebaruSan: That sounds cool! I don't think I have the requisite Docker expertise, though. Do you want to take this on?

HebaruSan: Also would that give us the option to leave the `tests` folder out of the production image?

techman83: That already happens as part of the setup.py build. It's why the production stage uses a --copy-from, to grab only the installation.

I think we should add a testing stage, but that would've taken longer than a coffee break to sort out. But I've done what will work for now and keeps mypy out of our prod image.

I didn't use the method I suggested, as that installs all the deps + the test one to global, which isn't ideal.

HebaruSan: Nice. I missed that the `COPY --from` allows us to install stuff that won't carry forward as long as it's not in `/home/netkan`. That solves this nicely.

techman83:
It's probably best to have new features/fixes in a separate PR, makes it easier to review. Overall looks good, so I'll merge it.

HebaruSan:
Well, I'm excited. This should catch most of the exceptions I kept causing before they make it into production.

HebaruSan:
Famous last words...

techman83:
> Well, I'm excited. This should catch most of the exceptions I kept causing before they make it into production.

That should be exceptions I also kept missing :wink:

PR #175, opened 2020-05-26
Check release date in AutoFreezer

HebaruSan:
## Problems

The AutoFreezer from #96 has itself been in cryofreeze due to mass-update changes:

- KSP-CKAN/CKAN#2760
- KSP-CKAN/CKAN#2922
- KSP-CKAN/CKAN#2942
- KSP-CKAN/CKAN#3061
- KSP-CKAN/NetKAN#7542 through KSP-CKAN/NetKAN#7598

It uses `ModStatus.last_indexed` to detect idle mods, so anytime we re-index a mod due to a change in how we generate metadata, we effectively fool the AutoFreezer into thinking that the mod was just updated. In reality we should be checking when the most recent release came out.

## Background

After KSP-CKAN/CKAN#3059, we will be capturing mods' release dates in metadata. They will pass through the Indexer on their way to CKAN-meta.

## Changes

- Now `Ckan` parses `release_date` into a `datetime` object
- Now `ModStatus` has a `release_date` property
- Now the Indexer copies `Ckan.release_date` into `ModStatus.release_date`
- Now the AutoFreezer's `_too_old` check looks at `release_date` if it's defined and falls back to `last_indexed` otherwise

This way we will be able to detect idle mods again.

### DIFF
--- a/netkan/netkan/metadata.py
+++ b/netkan/netkan/metadata.py
@@ -6,6 +6,7 @@ from hashlib import sha1
 import uuid
 import urllib.parse
 import github
+import dateutil.parser
 from typing import Optional, List, Tuple, Union, Any, Dict
 
 
@@ -253,6 +254,9 @@ class Ckan:
         'application/x-compressed-tar': 'tar.gz',
         'application/zip':              'zip',
     }
+    ISODATETIME_PROPERTIES = [
+        'release_date'
+    ]
 
     def __init__(self, filename: Union[str, Path] = None, contents: str = None) -> None:
         if filename:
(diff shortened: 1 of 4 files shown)


netkan/netkan/metadata.py
```diff
@@ -94,14 +95,27 @@ class Ckan:
...
+        self._raw = json.loads(self.contents, object_hook=self._custom_parser)
+
+    def _custom_parser(self, dct):
+        # Special handling for DateTime fields
+        for k in self.ISODATETIME_PROPERTIES:
+            if k in dct:
+                try:
+                    dct[k] = dateutil.parser.isoparse(dct[k])
+                except:
+                    pass
```
techman83, line 117: That's a pretty broad catch, probably ok here, but can lead to hiding errors we didn't expect.

HebaruSan: Yeah, that's intentional. Don't want the whole thing to come to a screeching halt because there's some problem parsing a date string. If that happens, it'll just leave the string in the dict, which is fine.

HebaruSan:
Waiting on KSP-CKAN/CKAN#3059, which is waiting on KSP-SpaceDock/SpaceDock#265 to go live in production (currently merged to alpha).

HebaruSan:
Those PRs are now merged, so we're about to have a big wave of `release_date` getting set. I want to watch how that goes before merging this.

PR #177, opened 2020-05-30
Webhook to generate metapackages for SpaceDock

HebaruSan:
## Motivation

SpaceDock has a modpack feature (internally called `ModList`), which doesn't do a lot today. Users can create a "pack" and add mods to it, but all this does is show those mods in one listing with separate Download buttons:

image

@V1TA5 is interested in adding functionality to make this more useful. One idea was to allow users to download a .ckan file for a modpack. SpaceDock can't do this by itself because it only has the SpaceDock id numbers for each mod, not the CKAN module identifiers.

## Changes

Now a new `/sd/metapackage` accepts a form list parameter of `mod_ids`, as well as form parameters for `name` and `abstract`, and returns JSON:

```json
{
    "missing": [ List of mod ids that are not in CKAN, in case SD wants to raise an error ],
    "metapackage": {
        Contents of .ckan file to install the modpack
    }
}
```

This way SpaceDock has the option to warn users and let them download a partial .ckan file if some modules are missing, or just cancel out.

Note that if multiple netkans share a SpaceDock entry, all will be added to the modpack.

If **no** modules are found, a 404 error is returned instead.

### DIFF
--- a/netkan/netkan/webhooks/__init__.py
+++ b/netkan/netkan/webhooks/__init__.py
@@ -10,6 +10,7 @@ from .errors import errors
 from .inflate import inflate
 from .spacedock_inflate import spacedock_inflate
 from .spacedock_add import spacedock_add
+from .spacedock_metapackage import spacedock_metapackage
 from .github_inflate import github_inflate
 from .github_mirror import github_mirror
 
@@ -28,6 +29,7 @@ class NetkanWebhooks(Flask):
         self.register_blueprint(inflate)
         self.register_blueprint(spacedock_inflate, url_prefix='/sd')
         self.register_blueprint(spacedock_add, url_prefix='/sd')
+        self.register_blueprint(spacedock_metapackage, url_prefix='/sd')
         self.register_blueprint(github_inflate, url_prefix='/gh')
         self.register_blueprint(github_mirror, url_prefix='/gh')
 
--- a/netkan/netkan/webhooks/spacedock_metapackage.py
+++ b/netkan/netkan/webhooks/spacedock_metapackage.py
@@ -1,0 +1,52 @@
+from pathlib import Path
+from flask import Blueprint, current_app, request, jsonify
+from typing import Tuple, Iterable, Dict, Any, List
+
+from ..common import pull_all
+from ..metadata import Netkan
+from ..repos import NetkanRepo
+
+
+spacedock_metapackage = Blueprint('spacedock_metapackage', __name__)  # pylint: disable=invalid-name
+
+
+# For metapackage hook on SpaceDock
+# Handles: https://netkan.ksp-ckan.space/sd/metapackage
+# POST form parameters:
+#     mod_ids:     The mods' ID numbers on SpaceDock
+@spacedock_metapackage.route('/metapackage', methods=['POST'])
+def metapackage_hook() -> Tuple[str, int]:
+    # Make sure our NetKAN and CKAN-meta repos are up to date
+    pull_all(current_app.config['repos'])
+    # Get the relevant netkans
+    sd_ids = request.form.getlist('mod_ids')
+    nks = find_netkans(current_app.config['nk_repo'], sd_ids)
+    if nks:
+        return jsonify({
+            'missing': not_found(sd_ids, nks),
+            'metapackage': mk_metapackage(nks, request.form.get('name', ''), request.form.get('abstract', '')),
+        }), 204
+    return 'No such modules', 404
+
+
+def mk_metapackage(netkans: Iterable[Netkan], name: str, abstract: str) -> Dict[str, Any]:
+    return {
+        'spec_version': 'v1.4',
+        'identifier': 'spacedock_metapackage',
+        'name': name,
+        'abstract': abstract,
+        'kind': 'metapackage',
+        'version': '1.0',
+        'license': 'unknown',
+        'depends': [{'name': nk.identifier} for nk in netkans]
+    }
+
+
+def find_netkans(nk_repo: NetkanRepo, sd_ids: List[str]) -> List[Netkan]:
+    return [nk for nk in nk_repo.netkans()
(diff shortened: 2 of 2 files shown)


techman83:
Hmm, I'm not 100% sure about this. If I understand correctly, it makes our hooks server an integral part of Spacedock's production workflow, with a function that isn't guaranteed to be reliable or fast. This could make troubleshooting a little hazy and potentially result in failures for end users.

Could Spacedock consume/cache our metadata to figure this out? We could add a webhook notification to let SD know when it's changed.

HebaruSan:
> Could Spacedock consume/cache our metadata to figure this out?

If I'm understanding this suggestion correctly, it would involve embedding the following knowledge in SpaceDock-owned code:

- Format of CKAN's metadata
- Structure of NetKAN repo
- The mapping from SpaceDock IDs to CKAN identifiers (which we just recently finished moving over to the CKAN side)
- Format of a metapackage

I think that all belongs in CKAN-owned code. It's implementation details of our stuff that shouldn't be tightly coupled to other projects.

So is the concern that we might slow down SpaceDock? Or that SpaceDock might overwhelm NetKAN's servers with extra traffic? I'm trying to strategize about how to start getting this back on track. I did work on KSP-SpaceDock/SpaceDock#293 recently, so modpacks should be less egregiously awful soon, but they could still use a "killer app" to make them really useful.

techman83:
> If I'm understanding this suggestion correctly, it would involve embedding the following knowledge in SpaceDock-owned code:
>
> Format of CKAN's metadata

The CKAN spec is known, consistent and well documented. It is designed to be consumable by any client.

> Structure of NetKAN repo

I'm not 100% across why this is needed?

> The mapping from SpaceDock IDs to CKAN identifiers (which we just recently finished moving over to the CKAN side)
> Format of a metapackage
> I think that all belongs in CKAN-owned code. It's implementation details of our stuff that shouldn't be tightly coupled to other projects.
>
> So is the concern that we might slow down SpaceDock? Or that SpaceDock might overwhelm NetKAN's servers with extra traffic?

This is a pretty big change from being an indexing service to an API provider used by end users. The service is _not_ architected with that in mind. If the webhooks go down for a short period, it's annoying, but it's not impacting anyone's ability to do the thing they want to do. Not only that, it's not rate limited, load balanced, redundant or multi-threaded. We'd probably want run a separate service, with it's own workers, own config. The webhooks also have limited but privileged access to our repos, github etc etc, which is locked down to only allow spacedock + github.

I'm not opposed to the concept as a whole. If as a project, we want to provide this service, we should make a conscious and deliberate decision to do so. We would want to step back and do some policy and technical design before throwing another endpoint into the webhooks.

Solid idea and I think it's a good one. Let's flesh it out and fully understand the requirements and consequences of running the service.

HebaruSan:
OK thanks, that makes sense. Pulling out the key points as notes to self for future reference:

- Rate limiting
- Load balancing
- Redundancy
- Multi-threading
- Separate service
  - Its own workers
  - Its own config

> The webhooks also have limited but privileged access to our repos, github etc etc, which is locked down to only allow spacedock + github.

For what it's worth, as I had imagined it, this would have continued to be the case. The user would have clicked a download link in SpaceDock, which would have gone to a route on SpaceDock's web server; _SpaceDock_, on the server side, would then have contacted the CKAN-owned server (whether it's the current NetKAN-Infra webhooks or some new service), received the result, and returned the result back to the user. I definitely am **not** proposing to open the NetKAN-Infra webhooks up to wide public access.

> If the webhooks go down for a short period, it's annoying, but it's not impacting anyone's ability to do the thing they want to do.

Continuing to brainstorm, did we discuss the possibility of doing this communication after a user saves a mod pack on SpaceDock, rather than on the fly at download? The SpaceDock database could add a `ModList.metapackage` column to store JSON received from NetKAN-Infra, update it in a background task when the mod pack changes, and then serve that up to users clicking the download link immediately. SpaceDock could manage a simple queue locally to re-try failed requests. This should keep the potential delay out of workflows and ensure that even a very popular mod pack would only need one request to CKAN. Does that address any of the concerns about reliability or load balancing?

techman83:
> OK thanks, that makes sense. Pulling out the key points as notes to self for future reference:
> 
> * Rate limiting
> * Load balancing
> * Redundancy
> * Multi-threading
> * Separate service
>   * Its own workers
>   * Its own config

Yep those are the important things to think about.

> For what it's worth, as I had imagined it, this would have continued to be the case. The user would have clicked a download link in SpaceDock, which would have gone to a route on SpaceDock's web server; _SpaceDock_, on the server side, would then have contacted the CKAN-owned server (whether it's the current NetKAN-Infra webhooks or some new service), received the result, and returned the result back to the user. I definitely am **not** proposing to open the NetKAN-Infra webhooks up to wide public access.
> 

Cool, that's probably a good approach for now. I would love to expand things more widely, but we'd have to do a whole lot of work to ensure things are secure, scale nicely and are reliable.

> Continuing to brainstorm, did we discuss the possibility of doing this communication after a user saves a mod pack on SpaceDock, rather than on the fly at download? The SpaceDock database could add a `ModList.metapackage` column to store JSON received from NetKAN-Infra, update it in a background task when the mod pack changes, and then serve that up to users clicking the download link immediately. SpaceDock could manage a simple queue locally to re-try failed requests. This should keep the potential delay out of workflows and ensure that even a very popular mod pack would only need one request to CKAN. Does that address any of the concerns about reliability or load balancing?

I think that's a pretty workable approach. Ultimately this is an in kind feature for SpaceDock and our service isn't geared up for on-demand end user things. We can always send rate-limiting responses if we needed to and SpaceDock could backoff/retry etc. I don't think we'd need to implement it on the first pass though. 

It would be nice if this were spun up as a separate container (we can use nginx to host it under a separate path like we did the legacy hooks), as we probably don't want to clash with the webhooks. But we're also almost at capacity on the instance.

PR #178, opened 2020-06-01
Refactor/repo interactions

techman83:
# Problem
We use the Repo class directly absolutely everywhere, along with lots of common operations. This makes considering switching git abstractions daunting.

This was a consequence of the organic development of an initial indexer replacement, to a massive improvement and increase in the number of things our automated indexing service is able to do. 

One of the big problems we currently have is our git interactions are IO heavy, it's hard to say if those are an issue with GitPython or how we're using it. Either way it's quite difficult to assess with the logic sprayed all over the place.

# Proposed Solution
This is very much an early iteration of what we can potentially do to work towards improving our interactions with git.

We have numerous common operations (list not exhaustive) that are common to how we use the repos:
* Commits
* Branching
* Push
* Pull

With the typing improvements, I notice we're returning objects only used in tests. Which in and of itself isn't a big deal, but something we ought to consider.

# Summary
This is a full refactor of the indexer git interactions, major changes as follows

* Indexer no longer accesses `Repo` directly
* `change_branch` context manager a function of `XkanRepo` with some documentation as it wasn't super clear about what it does
* Staged CKANs only branch when the inflation was successful and it involved a metadata change, this is a subtle but relatively major change in operation. Branch changing is a relatively expensive operation IO wise, so it really doesn't make sense to change branches when there is no change to the metadata. As an added bonus this cleans up `process_ckan`
* `download_counter` uses the commit function of CkanMetaRepo
* Some initial tests to ensure using a `nk_dir` / `ckm_dir` weren't broken in the change to a property
* Cleaned up some more of the linter warnings in the Indexer

### DIFF
--- a/netkan/netkan/download_counter.py
+++ b/netkan/netkan/download_counter.py
@@ -129,9 +129,8 @@ class DownloadCounter:
         )
 
     def commit_counts(self) -> None:
-        index = self.ckm_repo.git_repo.index
-        index.add([self.output_file.as_posix()])
-        index.commit(
+        self.ckm_repo.commit(
+            [self.output_file.as_posix()],
             'NetKAN Updating Download Counts'
         )
         logging.info('Download counts changed and commited')
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -5,11 +5,10 @@ from collections import deque
 from datetime import datetime, timezone
 from contextlib import contextmanager
 from dateutil.parser import parse
-from git import GitCommandError, Repo, Commit
-import boto3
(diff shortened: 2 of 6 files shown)


techman83:
> I think the part that consumed IO was contacting the remote to pull the branch. Nice to see that going away with the `metadata_changed` check.

Possibly, I'm not sure. If the if the repo is up to date, there isn't much IO. The old bot only staged when there were changes, so this is a bit of a return to that. Whilst retaining the simplicity of the current staging process.

> I assume we aspire to eventually eliminate all external access to the `.git_repo` properties?

That's absolutely the intention :slightly_smiling_face:

PR #181, opened 2020-06-02
Fix NetKAN hook freezer

### LINKED ISSUE #179: [Bug] GH Inflate Hook timing out
Background
----------
Github reports 503 errors and Gunicorn is timing out workers on `/gh/inflate`

Problem
-------
Freeze is likely time consuming as in dev it is slow with the small amount of objects in DynamoDB.
```python
def freeze(ids: List[str]) -> None:
    with ModStatus.batch_write() as batch:
        logging.info('Marking frozen mods...')
        for mod in ModStatus.scan(rate_limit=5):
            if not mod.frozen and mod.ModIdentifier in ids:
                logging.info('Marking frozen: %s', mod.ModIdentifier)
                mod.frozen = True
                batch.save(mod)
        logging.info('Done!')
```
This can likely be fixed with a more targeted DynamoDB query, but have disabled it in the interim to bring the service back.

HebaruSan:
The freezing feature from #173 was too slow and had to be disabled in #180.

Now instead of scanning all status records it just pull out the ones it needs. Should be way faster.

Fixes #179.

### DIFF
--- a/netkan/netkan/webhooks/github_inflate.py
+++ b/netkan/netkan/webhooks/github_inflate.py
@@ -88,8 +88,8 @@ def freeze(ids: List[str]) -> None:
                 status = ModStatus.get(ident)
                 if not status.frozen:
                     logging.info('Marking frozen: %s', ident)
-                    status.frozen = True
-                    status.save()
+                    # https://readthedocs.org/projects/pynamodb/downloads/pdf/stable/
+                    status.update(actions=[ModStatus.frozen.set(True)])
             except ModStatus.DoesNotExist:
                 # No status, don't need to freeze
                 pass
--- a/netkan/netkan/webhooks/github_inflate.py
+++ b/netkan/netkan/webhooks/github_inflate.py
@@ -26,6 +26,7 @@ def inflate_hook() -> Tuple[str, int]:
         current_app.logger.info('No commits received')
         return jsonify({'message': 'No commits received'}), 200
     inflate(ids_from_commits(commits))
+    freeze(frozen_ids_from_commits(commits))
     return '', 204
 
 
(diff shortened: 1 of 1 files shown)


netkan/netkan/webhooks/github_inflate.py
```diff
@@ -80,11 +81,16 @@ def frozen_ids_from_commits(commits: List[Dict[str, Any]]) -> List[str]:
 
 
 def freeze(ids: List[str]) -> None:
-    with ModStatus.batch_write() as batch:
+    if ids:
         logging.info('Marking frozen mods...')
```
techman83, line 85: 'ids' will always be a list, this if check here is unnecessary unless we specifically want to log the generic 'Marking mods frozen' - which is probably not all that useful.

netkan/netkan/webhooks/github_inflate.py
```diff
@@ -80,11 +81,16 @@ def frozen_ids_from_commits(commits: List[Dict[str, Any]]) -> List[str]:
...
+        for ident in ids:
+            try:
+                status = ModStatus.get(ident)
+                if not status.frozen:
+                    logging.info('Marking frozen: %s', ident)
+                    # https://readthedocs.org/projects/pynamodb/downloads/pdf/stable/
+                    status.update(actions=[ModStatus.frozen.set(True)])
+            except ModStatus.DoesNotExist:
+                # No status, don't need to freeze
+                pass
```
techman83, line 95: This looks good to me. If we ever try to freeze a few 100 mods at once this might fail, but that likely isn't something we need to worry about.

netkan/netkan/webhooks/github_inflate.py
HebaruSan: Yeah, the idea is that freezing a mod will be pretty rare (it happens, but probably <10%), so having the log full of "Marking frozen mods..." when nothing's getting frozen is not desirable.

PR #183, opened 2020-06-03
Set up SSH key before cloning repos in webhooks

HebaruSan:
The webhooks aren't starting.

```
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git clone --depth=1 -v git@github.com:KSP-CKAN/NetKAN.git /tmp/NetKAN
  stderr: 'Cloning into '/tmp/NetKAN'...
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

#182 refactored the webhooks' config info, but it moved the `init_ssh` call till after the repo clones. I could not remember what `init_ssh` was for or why we needed it, so I put it in `NetkanWebhooks.__init__` so it would be available during runtime. But we need it before that.

Now the ssh key is passed to `NetkanWebhooks.setup()`, so it can be stored before we clone.

### DIFF
--- a/netkan/netkan/webhooks/__init__.py
+++ b/netkan/netkan/webhooks/__init__.py
@@ -4,7 +4,6 @@ import boto3
 from flask import Flask
 from pathlib import Path
 
-from ..utils import init_ssh
 from ..notifications import setup_log_handler, catch_all
 from .config import current_config
 from .errors import errors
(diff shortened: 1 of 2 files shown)


techman83:
Great, thanks!!

PR #185, opened 2020-06-06
Fun with GraphQL (download counts)

HebaruSan:
## Motivation

```
$ curl -H "Authorization: token $GH_Token" https://api.github.com/rate_limit
{
  "resources": {
    "core": {
      "limit": 5000,
      "remaining": 3126,
      "reset": 1591429935
    },
    "search": {
      "limit": 30,
      "remaining": 30,
      "reset": 1591428892
    },
    "graphql": {
      "limit": 5000,
      "remaining": 5000,
      "reset": 1591432432
    },
    "integration_manifest": {
      "limit": 5000,
      "remaining": 5000,
      "reset": 1591432432
    },
    "source_import": {
      "limit": 100,
      "remaining": 100,
      "reset": 1591428892
    }
  },
  "rate": {
    "limit": 5000,
    "remaining": 3126,
    "reset": 1591429935
  }
}
```

This part represents API credits that we're "leaving on the table" since we currently don't use GraphQL at all:

```json
    "graphql": {
      "limit": 5000,
      "remaining": 5000,
      "reset": 1591432432
    },
```

Also GraphQL is neat and allows us to query more database objects in fewer calls, leaving out data we don't need.

image

## Changes

Now the download counts for GitHub are done via GraphQL. Mods hosted on GitHub are grouped into batches of 50 and then queried all at once. (I tried doing all of them in one gigantic request, but it timed out with an error, so we have to do smaller groups.)

- Fewer requests
- Much less total network traffic as only the download count data will be returned, without the unneeded info about repo names, owners, release tags, etc., etc.
- Parent, grandparent, etc. repos are retrieved without additional requests
- The separate `graphql` rate limiting pool is depleted instead of the main one for the API

Note that the old GitHub code is still needed to handle metanetkans that contain GitHub krefs. The batching only applies to netkans that go directly to GitHub without a metanetkan.

### DIFF
--- a/netkan/netkan/download_counter.py
+++ b/netkan/netkan/download_counter.py
@@ -5,6 +5,8 @@ from pathlib import Path
 from json.decoder import JSONDecodeError
 import requests
 import git
+from importlib.resources import read_text
+from string import Template
 from typing import Optional, Dict, Any
 
 from .metadata import Netkan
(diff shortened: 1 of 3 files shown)


netkan/netkan/download_counter.py
```diff
@@ -133,7 +241,7 @@ def commit_counts(self) -> None:
             [self.output_file.as_posix()],
             'NetKAN Updating Download Counts'
         )
-        logging.info('Download counts changed and commited')
+        logging.info('Download counts changed and committed')
         self.ckm_repo.git_repo.remotes.origin.push('master')
```
techman83, line 245: I can't believe I managed to persist misspelling 'committed' through 2 rewrites of this service :laughing:

netkan/netkan/download_counter.py
```diff
@@ -116,12 +211,25 @@ def __init__(self, nk_repo: NetkanRepo, ckm_repo: CkanMetaRepo, github_token: st
...
-                self.counts[netkanDl.identifier] = count
+            if netkanDl.kref_src == 'github':
+                # Process GitHub modules together in big batches
+                graph_query.add(netkanDl)
+                if graph_query.full():
+                    # Run the query
+                    graph_query.get_result(self.counts)
+                    # Start over with fresh query
+                    graph_query = GraphQLQuery(self.github_token)
+            else:
```
techman83, line 225: This will work perfectly and likely not a huge deal (being that it's very counter specific), but this logic should be owned by `GraphQLQuery`.

DasSkelett:
Looks like it did its job perfectly:
https://github.com/KSP-CKAN/CKAN-meta/commit/25dcc306d152fac8eb4cdbb9797d925893a41250

PR #188, opened 2020-06-19
Add GitHub Token Checks

techman83:
This mostly is a linting tidy up (it'll be nice when we can turn autoformat on without having it make changes to entire files!) and adds a sanity check so that we don't exhaust our github tokens.

The changes to the prod stack have already been applied, so the scheduler will be running every 30 minutes and has the token available to it for checking the limits.

### DIFF
--- a/netkan/netkan/common.py
+++ b/netkan/netkan/common.py
@@ -1,5 +1,7 @@
-from git import Repo
 from typing import List, Iterable, Dict
+
+from git import Repo
+import requests
 
 from .metadata import Netkan
 from .repos import NetkanRepo
@@ -24,3 +26,15 @@ def sqs_batch_entries(messages: Iterable[Dict[str, str]], batch_size: int = 10) 
 def pull_all(repos: Iterable[Repo]) -> None:
     for repo in repos:
         repo.remotes.origin.pull('master', strategy_option='theirs')
+
+
+def github_limit_remaining(token: str) -> int:
+    headers = {
+        'Authorization': f'token {token}',
+        'Content-Type': 'application/json'
+    }
+    response = requests.get(
+        'https://api.github.com/rate_limit',
+        headers=headers,
+    )
+    return response.json()['rate']['remaining']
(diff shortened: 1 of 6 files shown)


netkan/netkan/scheduler.py
```diff
@@ -91,8 +95,17 @@ def volume_credits_percent(self, cloudwatch: 'boto3.CloudWatch.Client', instance
             logging.error("Couldn't acquire Volume Credit Stats")
         return int(creds)
 
-    def can_schedule(self, max_queued: int, dev: bool = False, min_cpu: int = 50, min_io: int = 70) -> bool:
+    def can_schedule(self, max_queued: int, dev: bool = False,
+                     min_cpu: int = 50, min_io: int = 70, min_gh: int = 1500) -> bool:
         if not dev:
+            github_remaining = github_limit_remaining(self.github_token)
+            if github_remaining < min_gh:
```
HebaruSan, line 102: It would be nice if we could turn this check off for the SpaceDock webhooks-only pass, since it doesn't use the GitHub API. Maybe make `min_gh` a command line parameter?

netkan/netkan/common.py
```diff
@@ -24,3 +27,15 @@ def sqs_batch_entries(messages: Iterable[Dict[str, str]], batch_size: int = 10)
...
+def github_limit_remaining(token: str) -> int:
+    headers = {
+        'Authorization': f'token {token}',
+        'Content-Type': 'application/json'
+    }
+    response = requests.get(
+        'https://api.github.com/rate_limit',
+        headers=headers,
+    )
+    return response.json()['rate']['remaining']
```
HebaruSan, line 41: The `github` module's method for this is pretty nice!

```suggestion
    return github.Github(token).rate_limiting[0]
```

OR

```suggestion
    return github.Github(token).get_rate_limit().core.remaining
```

netkan/netkan/scheduler.py
techman83: That's a good idea, might leave for a future iteration. It shouldn't be a problem for now. It could use a refactor, we're stuffing a lot of logic and kwargs into that function!

netkan/netkan/common.py
techman83: Good catch! I forgot we were using that library, that's way nicer.

PR #189, opened 2020-06-20
Skip msg deletion if none succeed, print mirroring errors to Discord

HebaruSan:
## Problem

We're getting this every 5 minutes.

image

## Cause

The proximate cause is that `try_mirror` fails for SpaceTuxLibrary 0.0.3.1, so the list of messages to delete is empty, and boto3 thinks we care about that.

The ultimate cause is unknown; either the file isn't downloading, or there's a hash mismatch.

## Changes

Now we don't try to delete an empty list of messages. This will avoid the exception.

Now if `try_mirror` fails, the reason will be printed to Discord. This will allow us to continue investigating the root cause.


HebaruSan:
That answers that.

image

PR #190, opened 2020-06-20
Print more info about mismatched hashes

HebaruSan:
## Problem

After #189 we now get these every 5 minutes:

image

So now we know why it was failing before; the hashes don't match. There could be multiple reasons for that, maybe it's getting the wrong file out of the cache, or maybe a download actually failed. It will probably make sense to just delete the cached file and try again later, but I want to get more info about what's happening to make sure before we add that.

## Changes

Now in addition to the hashes, we will print which mod it is (already known but easier to keep track this way), and the path and size of the cached file. This will help us to determine why the hashes aren't matching.


HebaruSan:
Hoisted with my own petard.

image

It's the hash files!!

PR #191, opened 2020-06-20
Limit cache search to ZIP files

HebaruSan:
## Problem

After #190, we now get these every 5 minutes:

image

## Cause

The function that finds a mod's download in the cache is getting confused by the hash files from KSP-CKAN/CKAN#3080! It sometimes returns them instead of the actual download file.

## Changes

Now `Ckan.cache_find_file` only returns ZIP files.


HebaruSan:
image

PR #194, opened 2020-06-27
Purge and replace cached downloads in Mirrorer

HebaruSan:
## Problem

We now get this every 5 minutes:

image

## Cause

In KSP-CKAN/CKAN-meta#2020, we updated this mod's hashes because its download was replaced after being uploaded. But now the Mirrorer is confused because the original file is still in the cache and doesn't match the metadata. Currently all the Mirrorer can do in this situation is print an error and bail.

## Changes

- `large_file_sha256` is moved from `Mirrorer` to `CkanMirror`
- A new `CkanMirror.open_if_hash_match` function checks the hash and deletes the file if it doesn't match
- `CkanMirror.open_download()` now checks the hashes for both the cached file and the fresh download, if any

This will allow the Mirrorer to replace an out of date download in the cache.

### DIFF
--- a/netkan/netkan/mirrorer.py
+++ b/netkan/netkan/mirrorer.py
@@ -207,22 +207,60 @@ class CkanMirror(Ckan):
             **({'licenseurl': lic_urls} if lic_urls else {}),
         }
 
+    @staticmethod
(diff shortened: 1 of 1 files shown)


HebaruSan:
:crossed_fingers:

PR #195, opened 2020-07-05
List frozen mods in AutoFreezer pull requests

HebaruSan:
## Motivation

Handling KSP-CKAN/NetKAN#8004 was a bit tedious; it initially froze 251 mods, 6 of which were unfrozen before commit. The list of mods was only available in the commit list, and the release dates were only available via external means such as the status page.

## Changes

Now the AutoFreezer's pull requests will include a two-column table showing each frozen mod and its last release datetime / last indexed datetime, sorted by the timestamp. This will make it easier to see what is being frozen and why.


techman83:
Nice improvement!

PR #198, opened 2020-11-19
Drop mods from auto-freeze list after 3 weeks

HebaruSan:
## Background

The AutoFreezer finds all unfrozen mods that haven't been updated in 1000+ days and submits a pull request to freeze them all, see KSP-CKAN/NetKAN#8233. We can then review the list to determine which ones should really be frozen and which ones are likely to have new releases or version file updates or otherwise remain active.

## Problem

We're accumulating a long-ish list of mods that we don't actually want to freeze. Two examples:

- SpaceLink is a SpaceDock-related mod, which has its compatibility updated every time there's a new KSP version, for a release from April 2017. If we freeze it, we miss those upates.
- PlaneMode was originally frozen in #8004, then unfrozen in https://github.com/KSP-CKAN/NetKAN/commit/bcc7016b533ea7866386f843134cbcdb914d6eea because it's still compatible

Once we decide we don't need to freeze a mod, it's not necessary to keep reviewing that decision every week.

## Changes

Now the list only includes mods that were last updated 1000-1021 days ago. This gives us a 3-week window (and 3 pull requests) to decide whether to freeze a mod, after which it drops off the list. As a result the lists in the pull requests will be shorter and more relevant.

The limit shouldn't be 1 week, because the last time I took a break from doing CKAN stuff, we had an AutoFreezer PR sit around for a few weeks, blocking others from being submitted. If that happened again, we could miss our chance to freeze any mods that expired in that time.

### DIFF
--- a/netkan/netkan/auto_freezer.py
+++ b/netkan/netkan/auto_freezer.py
@@ -17,9 +17,9 @@ class AutoFreezer:
         self.nk_repo = nk_repo
         self.github_pr = github_pr
 
-    def freeze_idle_mods(self, days_limit: int) -> None:
+    def freeze_idle_mods(self, days_limit: int, days_till_ignore: int) -> None:
         self.nk_repo.git_repo.remotes.origin.pull('master', strategy_option='ours')
-        idle_mods = self._find_idle_mods(days_limit)
+        idle_mods = self._find_idle_mods(days_limit, days_till_ignore)
         if idle_mods:
             self._checkout_branch(self.BRANCH_NAME)
             for ident, _ in idle_mods:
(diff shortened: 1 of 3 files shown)


HebaruSan:
I think the build error is complaining that `key` can return `None` here (because `Ckan.version` can return `None`):

https://github.com/KSP-CKAN/NetKAN-Infra/blob/357b9dd2945737ab0f0195c29b7e10efa96be843/netkan/netkan/repos.py#L155-L157

https://github.com/KSP-CKAN/NetKAN-Infra/blob/357b9dd2945737ab0f0195c29b7e10efa96be843/netkan/netkan/metadata.py#L286-L290

```python
>>> max(1,2,3)
3
>>> max(1,2,3,None)
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

Nice catch by mypy. Will look into it.

HebaruSan:
```
netkan/repos.py: note: In member "highest_version" of class "CkanMetaRepo":
netkan/repos.py:156:19: error: Value of type variable "_LT" of "max" cannot be
"Version"  [type-var]
            highest = max(self.ckans(identifier), default=None, key=lambda...
                      ^
```

HebaruSan:
Seems to be a change in the latest mypy, 0.790. My local copy of 0.780 still says there are no issues.

techman83:
Yep, seems reasonable to me.

PR #200, opened 2020-11-26
Move Project to GitHub Actions

techman83:
Travis has served us well for some time, however GitHub Actions are far more flexible and are very very fast.

# Adds
- mypy testing available for any branch push
- unit test for any branch push

# Changes
- Container builds happen within GitHub Actions
- Service redeployment happen within GitHub Actions

# Removes
- Travis Definition
- Deployment script for travis


.github/workflows/build.yml
```diff
@@ -0,0 +1,24 @@
...
+    types: [opened, synchronize, reopened]
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
```
HebaruSan, line 14: I thought QEMU was a machine emulator like VMware. Why do we need it to build NetKAN-Infra?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,72 @@
+name: Deploy NetKAN-Infra
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    if: ${{ github.repository == 'KSP-CKAN/NetKAN-Infra' }}
```
HebaruSan, line 10: Might be better to make this check whether the secrets are defined, like the one for CKAN does:

https://github.com/KSP-CKAN/CKAN/blob/e39d735583e2e965cc99e8196e8f76a99ea3c822/.github/workflows/deploy.yml#L71

```yml
if: ${{ env.DOCKERHUB_USERNAME && env.DOCKERHUB_PASSWORD }}
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: ./nginx
+          file: ./nginx/Dockerfile
+          push: false
+          tags: kspckan/webhooks-proxy:latest
+
+  deploy:
+    if: ${{ github.repository == 'KSP-CKAN/NetKAN-Infra' }}
```
HebaruSan, line 51: Similar comment as for above `if` statement:

```yml
if: ${{ env.AWS_ACCESS_KEY_ID && env.AWS_SECRET_ACCESS_KEY }}
```

.github/workflows/build.yml
DasSkelett: It is included in the usage examples for the Docker build-push action:
See https://github.com/marketplace/actions/build-and-push-docker-images#usage

Not sure if it's actually a hard dependency for this buildx or not. They talk about it being used for "multi-platform" images in the [README](https://github.com/docker/buildx#building-multi-platform-images).

HebaruSan: OK, those links make it sound like we don't need QEMU...

- We only need to build for one arch (whatever we're using for the bot)
- It's just Python anyway
- We don't need to run binaries from other architectures

Is there a way I can test this locally with those lines removed, to confirm?

techman83: Yeah, fair points. I'll try it, the documentation wasn't super clear about the point.

.github/workflows/deploy.yml
techman83: I'm fine with being consistent across the projects.

.github/workflows/build.yml
techman83: Seems to work, I'll remove all the QEMU statements. Saves us 10 seconds on each build :smile:

HebaruSan (approved):
I'm ready to try this out if you are!

DasSkelett:
I for one would've probably put mypy and the unit tests into one file and job, so we only have to download and setup Python and our dependencies once, but that's personal preference I guess.

HebaruSan:
> put mypy and the unit tests into one file and job, so we only have to download and setup Python and our dependencies once

Should we do that for SpaceDock, @DasSkelett?

techman83:
> I for one would've probably put mypy and the unit tests into one file and job, so we only have to download and setup Python and our dependencies once, but that's personal preference I guess.
> 
> Otherwise this looks fine to me, too. Thanks a ton!

I think the nice part here is that it is obvious what failed, pretty quick to run, and run concurrently.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          context: ./nginx
+          file: ./nginx/Dockerfile
+          push: false
+          tags: kspckan/webhooks-proxy:latest
+
+  deploy:
+    if: ${{ env.AWS_ACCESS_KEY_ID && env.AWS_SECRET_ACCESS_KEY }}
+    runs-on: ubuntu-latest
+    needs: build
```
DasSkelett, line 52: Speaking of concurrency, do we need to add `webhooks` here`? We are restarting the `Webhooks` service below and if I read the `prod-stack.py` correctly, that one also needs the `kspckan/webhooks-proxy` image from the `webhooks` job above.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+      - name: Build and push webhooks-proxy
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: ./nginx
+          file: ./nginx/Dockerfile
+          push: false
```
DasSkelett, line 46: Also not 100% sure, but I think this should be `true`? We don't seem to be building and pushing that image anywhere else.

.github/workflows/deploy.yml
techman83: Good catch!

.github/workflows/deploy.yml
techman83: Yes, copy/paste fail :grimacing:

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+  webhooks:
+    if: ${{ secrets.DOCKERHUB_USERNAME && secrets.DOCKERHUB_PASSWORD }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
```
DasSkelett, line 39: This probably also needs the credentials like above.

techman83: Fffffffff! Really need a :coffee:

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+          file: ./nginx/Dockerfile
+          push: true
+          tags: kspckan/webhooks-proxy:latest
+
+  deploy:
+    if: ${{ env.AWS_ACCESS_KEY_ID && env.AWS_SECRET_ACCESS_KEY }}
+    runs-on: ubuntu-latest
+    needs: [ 'build', 'webhooks' ]
+    steps:
+    - uses: actions/checkout@v1
```
DasSkelett, line 54: Here and in `mypy.yml` and `unittest.yml` we are using `actions/checkout@v1`, whereas everywhere above we are using `actions/checkout@v2`. Probably doesn't make a difference, but it would be nice to have stay consistent.

.github/workflows/deploy.yml
DasSkelett: Haha, no worries. I also didn't spot these the first couple of times I've read looked at it.

.github/workflows/deploy.yml
techman83: Yeah, those were lifted straight from @HebaruSan 's examples. I'll line them up.

DasSkelett (approved):
Okay, that's it for good. Well, it'll tell us otherwise.
Thanks a lot again!

techman83:
Awesome, thanks for helping make it nice and tidy!

PR #203, opened 2020-12-18
Webhooks Memory Allocation

techman83:
Problem
=======
The webhooks service kept failing with
```
git.exc.GitCommandError: Cmd('git') failed due to: exit code(-9)
```
This is likely an OOM.

Change
======
I've bumped the memory for the service and we can monitor it. This has already been applied to the NetKAN stack.


DasSkelett (approved):
Time will tell :P

PR #204, opened 2020-12-29
Add permissions for CI

techman83:
CI Required permission to add items to the queue, this has been applied already.

### DIFF
--- a/prod-stack.py
+++ b/prod-stack.py
@@ -478,6 +478,36 @@ t.add_resource(PolicyType(
     }
 ))
 
+# Metadata CI Permissions
+# CI access for metadata actions
+ksp_ci_metadata_group = t.add_resource(Group("KspCkanCiMetadataGroup"))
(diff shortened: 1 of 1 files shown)


HebaruSan:
So, I have no idea what this does...
it wasn't working? How does this tie to the new Action in a way it didn't previously? It looks like this just allows listing queues?

techman83:
> So, I have no idea what this does...
> it wasn't working? How does this tie to the new Action in a way it didn't previously? It looks like this just allows listing queues?

I created a new user specifically for the Metadata CI and added it to the 'KspCkanCiMetadataGroup' permission group. It needs to be able to list the queues (to get the queue url) and add a message to the queue.

HebaruSan (approved):
OK, that makes sense. I suppose we'll find out if there are any problems!

techman83:
Haha, it should be relatively pretty binary. Works or doesn't work :slightly_smiling_face:

PR #205, opened 2021-01-06
Store cache timestamps to status pynamodb

HebaruSan:
## Motivation

Currently the `ModStatus` objects have properties `last_checked` and `last_inflated`, which are identical, which is redundant. This is not a coincidence; the code sets them to the exact same value every time. They also appear on the status page:

- http://status.ksp-ckan.space/

image

## Problems

In https://github.com/KSP-CKAN/CKAN-meta/commit/8006736930cc1bcf595afe2cd86faea3916186cb, KerbalWeatherProject's release date was updated but its hashes were not. Subsequently attempting to install this mod resulted in hash mismatches:

```
CKAN.InvalidModuleFileKraken: KerbalWeatherProject v1.0.0: C:\Users\aequa\AppData\Local\Temp\tmp83D1.tmp has length 88819967, should be 88820268
   at CKAN.NetModuleCache.Store(CkanModule module, String path, String description, Boolean move)
   at CKAN.NetAsyncModulesDownloader.ModuleDownloadComplete(Uri url, String filename, Exception error)
```

So the author replaced the download, but the Inflator used the old download to re-index it.

## Causes

There are two ways the Inflator might have inflated the old file:

- The Inflator thought its cached download was up to date and didn't re-download (i.e., KSP-CKAN/CKAN#3246 is broken or did not fully fix the problem)
- The Inflator requested to re-download but SpaceDock sent the old file (i.e., KSP-SpaceDock/SpaceDock#313 is broken or too slow)

Right now we cannot tell which it is, because we do not know the timestamp of the file in the Inflator's cache. We need that info to continue investigating and eventually find a fix. If the timestamp is close to the re-indexing request, then we blame SpaceDock for lying, otherwise we blame the Inflator for laziness.

## Changes

Now `last_checked` is no longer set (but we won't purge the current values, just in case). `last_inflated` will still be set to their common value, so no information is lost.

Now a new `last_downloaded` property is set to the timestamp of the mod's download in the cache. To make this possible, the Indexer now mounts the cache directory, finds downloads in it, and `stat`s them to get the timestamp. We carefully convert this to UTC to ensure that we can always track exactly what happened with a mod, minute by minute.

A future PR will update the status page to show the new timestamps.

### DIFF
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -88,17 +88,11 @@ class CkanMessage:
         return commit
 
     def status_attrs(self, new: bool = False) -> Dict[str, Any]:
-        inflation_time = parse(self.CheckTime)
         attrs: Dict[str, Any] = {
             'success': self.Success,
             'last_error': self.ErrorMessage,
             'last_warnings': self.WarningMessages,
-            # We may wish to change the name in the inflator
-            # as the index will set 'last_checked'
-            'last_inflated': inflation_time,
-            # If we have perfomed an inflation, we certainly
-            # have checked the mod!
-            'last_checked': inflation_time,
+            'last_inflated': parse(self.CheckTime),
             # If we're inflating it, it's not frozen
             'frozen': False,
         }
@@ -107,6 +101,10 @@ class CkanMessage:
             attrs['resources'] = resources
         if new:
             attrs['ModIdentifier'] = self.ModIdentifier
+        cache_path = self.ckan.cache_find_file
+        if cache_path:
+            cache_mtime = datetime.fromtimestamp(cache_path.stat().st_mtime)
+            attrs['last_downloaded'] = cache_mtime.astimezone(timezone.utc)
         if self.indexed:
             attrs['last_indexed'] = datetime.now(timezone.utc)
         release_date = getattr(self.ckan, 'release_date', None)
--- a/netkan/netkan/status.py
+++ b/netkan/netkan/status.py
@@ -40,6 +40,7 @@ class ModStatus(Model):
     last_checked = UTCDateTimeAttribute(null=True)
     last_indexed = UTCDateTimeAttribute(null=True)
     last_inflated = UTCDateTimeAttribute(null=True)
+    last_downloaded = UTCDateTimeAttribute(null=True)
     release_date = UTCDateTimeAttribute(null=True)
     success = BooleanAttribute()
     frozen = BooleanAttribute(default=False)
--- a/netkan/tests/indexer.py
+++ b/netkan/tests/indexer.py
@@ -109,7 +109,6 @@ class TestCkan(unittest.TestCase):
         attrs = self.message.status_attrs(new=True)
         self.assertEqual(attrs['ModIdentifier'], 'DogeCoinFlag')
         self.assertTrue(attrs['success'])
-        self.assertIsInstance(attrs['last_checked'], datetime)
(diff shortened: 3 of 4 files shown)


HebaruSan:
Oops, that's right, pynamodb isn't just a free-form JSON blob, you have to define the columns. One moment please...

techman83:
Yep, 100% agree. The 'last_checked' field was an artifact of unnecessary future plans that never came to fruition. Realistically with all the improvements that have been made to the inflator, those ideas are no longer relevant. Love your work!

netkan/netkan/indexer.py
```diff
@@ -107,6 +101,10 @@ def status_attrs(self, new: bool = False) -> Dict[str, Any]:
             attrs['resources'] = resources
         if new:
             attrs['ModIdentifier'] = self.ModIdentifier
+        cache_path = self.ckan.cache_find_file
+        if cache_path:
+            cache_mtime = datetime.fromtimestamp(cache_path.stat().st_mtime)
+            attrs['last_downloaded'] = cache_mtime.astimezone(timezone.utc)
```
techman83, line 107: My only note here is this will involve IO, require exposing the cache path to the indexer, and relying on file system timestamps. Which whilst not a big deal, the Inflator already knows when something is a cache hit or miss. Could we add this as a field on our SQS message? Is this significant effort on the inflator side of things?

techman83 (changes requested):
Really only a minor implementation detail here, if effort is significant, filesystem timestamps will be sufficient.

netkan/netkan/indexer.py
HebaruSan: > involve IO

Well, for each mod, we'll have to:

1. Find the file
   Uses `Path.glob` to search the cache folder by filename. Unix filesystems list the file names contained in a directory in a dense sequence of contiguous lists (as many as will fit in a block), so the locality of reference here would be very high, one disk hit for a large number of file names (we might have around 3000-4000?). It doesn't load anything specific to the file until you ask for something beyond just the name. And these blocks are likely to be and stay cached.
   http://www.porcupine.org/forensics/forensic-discovery/chapter3.html#figure3.2
   image
2. Get the timestamp
   `stat`s the file and retrieves the mtime from the stat struct. I think this will be one disk access, for just the one file we found in the previous step.

So I don't expect a significant hit from this; it should be trivial next to all the other work the Infra does. I can look into gathering some measurements if you think that would be wise, though.

> require exposing the cache path to the indexer

Yup, I updated `prod-stack.py` to hopefully do this. Does that look like it will work?

> relying on file system timestamps

That's definitely what I want. ;)

> the Inflator already knows when something is a cache hit or miss

Hmm, I'd like to be able to backfill these values reasonably as well, which this PR's current code will handle for free by checking the currently cached files. The Inflator now has no idea when it last had to download each mod.

> Could we add this as a field on our SQS message? Is this significant effort on the inflator side of things?

There may be a technical quirk there. Netkan.exe has multiple places where it _might_ download a given file, and it checks the cache first each time (basically each translator or validator that wants to look at the ZIP asks the `CachingHttpService` to provide the file, independently, and it relies on the cache to avoid doing this repeatedly). So the first inflation of a new module would look like one cache miss followed by several cache hits. It feels like it would get messy to try to "remember" when those misses happen and when to consider them no longer applicable. (But I see reasons to avoid this approach anyway as discussed in the previous points.)

techman83: I think that's a reasonable argument, we can always cross that bridge if we need to at a later date.

PR #209, opened 2021-01-28
Ensure Discord minimum

techman83:
This will ensure we get the version with aiohttp with the fixed deps.


HebaruSan:
```
ERROR: Could not find a version that satisfies the requirement discord>=1.6.0 (from netkan-indexer[test])
ERROR: No matching distribution found for discord>=1.6.0
```

DasSkelett (approved):
Yup, let's give it a try

Issue #208, opened 2021-01-28
[Bug] Indexing services fail to start


techman83:
Background
----------
NetKAN Production

Problem
-------
Upstream discord lib `aiohttp` is missing `typing_extensions`, noted and fixed in aio-libs/aiohttp#5107. However in production we still seem to have caught version missing the dep.

```
Traceback (most recent call last):
  File ".local/bin/netkan", line 5, in <module>
    from netkan.cli import netkan
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/__init__.py", line 3, in <module>
    from .services import (
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/services.py", line 6, in <module>
    from .common import common_options, pass_state, SharedArgs
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/common.py", line 10, in <module>
    from ..notifications import setup_log_handler, catch_all
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/notifications.py", line 6, in <module>
    import discord
  File "/home/netkan/.local/lib/python3.7/site-packages/discord/__init__.py", line 25, in <module>
    from .client import Client
  File "/home/netkan/.local/lib/python3.7/site-packages/discord/client.py", line 33, in <module>
    import aiohttp
  File "/home/netkan/.local/lib/python3.7/site-packages/aiohttp/__init__.py", line 6, in <module>
    from .client import (
  File "/home/netkan/.local/lib/python3.7/site-packages/aiohttp/client.py", line 35, in <module>
    from . import hdrs, http, payload
  File "/home/netkan/.local/lib/python3.7/site-packages/aiohttp/http.py", line 7, in <module>
    from .http_parser import (
  File "/home/netkan/.local/lib/python3.7/site-packages/aiohttp/http_parser.py", line 15, in <module>
    from .helpers import NO_EXTENSIONS, BaseTimerContext
  File "/home/netkan/.local/lib/python3.7/site-packages/aiohttp/helpers.py", line 48, in <module>
    from typing_extensions import Protocol
ModuleNotFoundError: No module named 'typing_extensions'
```

PR #211, opened 2021-01-28
Fix dependency issues

techman83:
This PR shifts the `--help` sanity check and will add (testing for failure first) a `--force-reinstall` to ensure all deps get installed to user.


techman83:
That fixes the underlying issue of #208 and also moves our sanity check to avoid it happening again in the future.

Issue #210, opened 2021-01-28
[Bug] System 'mypy' installation in base causes missing deps in prod


techman83:
Problem
-------
We use a mutlistage `Dockerfile` to reduce image deployment size and deployment times. Any test only dependencies get installed to system prior to installing netkan to user.

`mypy` now includes `typing-extensions`, when we install as a user, it sees the system package. When we copy the user install to the production stage, that dependency is no longer present.

```
Requirement already satisfied: typing-extensions>=3.6.5 in /usr/local/lib/python3.7/site-packages (from aiohttp<3.8.0,>=3.6.0->discord.py>=1.6.0->netkan-indexer==1.0) (3.7.4.3)
```

Fix
------
- Move `--help` sanity test to production image step
- Use `--force-reinstall`

Issue #49, opened 2019-10-03
[Project] Style Guide


techman83:
Problem
-------
Decoupling the project has made it a lot easier to be contributed to. But also means with multiple people working on it, time is being spent on style consistency rather than code.

Suggestions
-----------
While I don't necessary think we should ever hold up code that makes things better, a style guide will make it easier to maintain consistency. So that new and existing developers can get underway with code changes more quickly.

We can probably add some automated linting/testing at the same time. Let the machines do the hard work!

techman83:
#214 will ensure we have some form of consistency, so I'm happy enough with this.

PR #212, opened 2021-01-28
Unified testing framework

HebaruSan:
## Problem

See #210, installing mypy in the Dockerfile caused problems related to dependencies-of-dependencies that were installed at the system level and then not available on a fresh image with only the user install copied over.

## Motivation

In the Discord discussion of the first iteration of this pull request, we agreed that it would be nice to have one simple command that developers could run to test everything, and to re-use that same command in our GitHub workflows. It looked like the `test_suite` option in `setup.py`, supporting the `python setup.py test` command, might be a step in that direction.

Enter pytest-dev/pytest#5534. They're deprecating `python setup.py test`.

In the Discord discussion of the _second_ iteration of this pull request, `tox` was ruled out as a solution because we already use Docker to manage our virtual environments and would prefer to run our tests in the same environment where the actual code runs. So now `pytest` is winning.

## Changes

- Now we don't install or run mypy in the Dockerfile
- A new `test` workflow is created to run all tests
- The separate mypy and unittest workflows are deleted

For me, this is how to run tests, though if you use a distro that doesn't try to mix Python 2 and 3 like Ubuntu does, drop the `-3`:

```
cd netkan
pip install --user .[test]
pytest-3 --mypy -v
```

This provides a single uniform access point for all testing, which will work the same for developers and for CI/CD.


HebaruSan:
I guess unittest is not a package?

techman83:
> I guess unittest is not a package?

`unittest` is a part of python standard lib.

.github/workflows/mypy.yml
```diff
@@ -14,8 +16,8 @@ jobs:
         python-version: 3.7
     - name: Install Dependencies
       run: |
-        pip install netkan/.[test]
+        pip install --user netkan/. --no-warn-script-location --force-reinstall
+        pip install mypy
```
techman83, line 20: Our test dependencies *should* come from within `setup.py`. The reason pip was installing mypy directly was that calling `[test]` will install all the deps plus the test ones. `--force-reinstall` ensures all deps get installed as user, there isn't any need to install as user in this environment.

.github/workflows/unittest.yml
```diff
@@ -14,8 +16,7 @@ jobs:
         python-version: 3.7
     - name: Install Dependencies
       run: |
-        pip install netkan/.[test]
+        pip install --user netkan/. --no-warn-script-location --force-reinstall
     - name: Run unittest
```
techman83, line 20: See previous comment.

netkan/Dockerfile
```diff
@@ -3,12 +3,9 @@ RUN useradd -ms /bin/bash netkan
 ADD . /netkan
 WORKDIR /netkan
 RUN pip install pip --upgrade
-RUN pip install mypy
 RUN chown -R netkan:netkan /netkan
 USER netkan
 RUN pip install --user . --no-warn-script-location --force-reinstall
-RUN mypy .
-RUN python -m unittest -v
```
techman83, line 9: Our docker environment isn't our github one, we likely still want to run our tests inside the container.

netkan/setup.py
```diff
@@ -38,9 +38,6 @@
             'pylint',
             'autopep8',
             'troposphere',
-        ],
-        'test': [
-            'mypy',
         ]
```
techman83, line 41: Our package should be explicit about its test dependencies.

techman83 (commented):
I think we can drop `mypy` from the container, it's non-runtime type checking. But our tests should still be run inside the container.

Though we don't use `pytest` or any other test packages in the future, so we may want to bust out `mypy` into a separate dependency set.

.github/workflows/mypy.yml
HebaruSan: I don't think we've been well-served by the concept of "test dependencies". We need mypy here and nowhere else, and installing it elsewhere caused 10 hours of downtime.

netkan/Dockerfile
HebaruSan: Makes sense to me.

netkan/setup.py
HebaruSan: Mypy isn't needed to run unittest. I think we should handle such things individually, so we only have the tools we need in each case.

techman83: `mypy` is technically a test dependency, but you are correct it isn't a `unittest` dependency. However if we say move to `pytest`, we would actually have more than one dependency for testing.

HebaruSan: But they would still be separate dependencies for separate workflows. What's the benefit in combining them into one big list and installing extra stuff?

.github/workflows/mypy.yml
techman83: Well, it was the manual installation of `mypy` in the container that caused the downtime. We won't be installing our test dependencies in the container for now and if we do, we'll need to think about how best to approach it. For example we could have a separate 'test' stage, that is `FROM production as test` that we could run as an build action

```Dockerfile
FROM production as test
USER root
ADD . /netkan
RUN chown -R netkan:netkan /netkan
WORKDIR /netkan
USER netkan
RUN pip install --user /netkan/.[test]
RUN mypy .
RUN pytest -v
```


```yml
      - name: Build Test
        id: docker_build
        uses: docker/build-push-action@v2
        with:
          context: ./netkan
          file: ./netkan/Dockerfile
          push: false
          target: test
```

netkan/setup.py
techman83: The benefit is that if you are someone who wants to run our tests, that you get all the dependencies required for running said tests. What we do in CI/CD we can document in code; but if a python dev comes along with a fix, they should be able to go `pip install .[test]` and have everything they require to run them.

It's more about our package being correct than being helpful for ci. Though we could add a separate stanza for ci only packages if we want to be clear about it.

.github/workflows/unittest.yml
techman83: We actually don't use any dependencies for `unittest` - we can adjust that later if that changes.

netkan/setup.py
HebaruSan: Hmm, mypy isn't a test, though, in the sense you're talking about.

HebaruSan:
It's not running the old workflows or the new one. Clearly I've missed something. Any tips appreciated.

netkan/setup.py
HebaruSan: The new `tox.ini` file specifies the test dependencies in the `deps =` line.

techman83 (commented):
Tox has its use case, but I feel it's a little redundant. We can already manage environments in our workflows and our container.

`pytest` is the go to for testing frameworks. It has wrappers for mypy, linting, flake8 etc. Which would allow us to close off #49

It is rather daft that the `python setup.py test` is now deprecated, but the python community have been re-inventing the packaging wheel for as long as I can remember.

techman83:
Ok, so I've put my code where my mouth is! :slightly_smiling_face: 

* `test` stage for docker, which uses production as a base. These layers won't end up in production as we target `production`, but will test in the same environment.
* keeping the `--help` check in production is still a good sanity check
* pytest used in the test workflow
* tests are run before the deployment is produced
* vs code environment works correctly (this was a right pain..)

HebaruSan:
What's the command for running tests locally now? I'm not finding one that works yet:

```
$ pytest
Command 'pytest' not found, did you mean:
  command 'ptest' from deb pbuilder-scripts (22)
  command 'ptest' from deb parmetis-test (4.0.3-5build1)
  command 'yptest' from deb nis (3.17.1-5)
Try: sudo apt install <deb name>

$ py.test-3 --mypy -v
ERROR: usage: py.test-3 [options] [file_or_dir] [file_or_dir] [...]
py.test-3: error: unrecognized arguments: --mypy
  inifile: /home/user/Downloads/KSP/NetKAN-Infra/netkan/pytest.ini
  rootdir: /home/user/Downloads/KSP/NetKAN-Infra/netkan
```

HebaruSan:
@techman83 your changes look good to me. Should we look into pylint and flake8?

techman83:
Yes, but I think a separate PR. I enabled pylint and got a whole suite of warnings I didn't want to deal with on a PR for unifying our testing.

PR #213, opened 2021-01-28
Omit step IDs

techman83:
Deployment build failed with the following:
```
The workflow is not valid. .github/workflows/deploy.yml (Line: 31, Col: 13): The identifier 'docker_build' may not be used more than once within the same scope.
```
We don't use the outputs from these steps, so we don't require them.


HebaruSan (approved):
:+1:

PR #215, opened 2021-03-23
Pull shallowly and collect garbage

HebaruSan:
## Problem

We're seeing occasional/periodic issues where the volume credits for the bot plummet sharply:

graph

## Cause

We don't really know; there just isn't a lot of evidence at this point. One consensus possibility is that it has to do with git and/or gitpython getting slower and/or using more RAM over time as more and more commits accumulate in the history.

(I'm privately skeptical of this explanation, as I work with plenty of full-depth clones of plenty of repos and have not seen slowness or memory issues in working with them, and git is pretty well designed in general to avoid rookie mistakes. However, I also can't rule it out.)

## Changes

Now our `git pull` commands use `--depth=1` to keep the clone shallow. `--allow-unrelated-histories` is needed to merge the remote into the local master without a common base commit.
Then we `git gc --prune=all` to remove dangling objects.

This should eliminate all accumulation of complex data in git. If the de-shallowing of our clones is a cause, then this will fix the problems. If it isn't, then now we will know that for sure.

### DIFF
--- a/netkan/netkan/common.py
+++ b/netkan/netkan/common.py
@@ -27,7 +27,8 @@ def sqs_batch_entries(messages: Iterable[Dict[str, str]],
 
 def pull_all(repos: Iterable[Repo]) -> None:
     for repo in repos:
-        repo.remotes.origin.pull('master', strategy_option='theirs')
+        repo.remotes.origin.pull('master', strategy_option='theirs', depth='1', allow_unrelated_histories=True)
+        repo.git.gc(prune='all')
 
 
 def github_limit_remaining(token: str) -> int:
--- a/netkan/netkan/mirrorer.py
+++ b/netkan/netkan/mirrorer.py
@@ -322,7 +322,8 @@ class Mirrorer:
                 # Get up to date copy of the metadata for the files we're mirroring
                 logging.info('Updating repo')
                 self.ckm_repo.git_repo.heads.master.checkout()
-                self.ckm_repo.git_repo.remotes.origin.pull('master', strategy_option='theirs')
+                self.ckm_repo.git_repo.remotes.origin.pull('master', strategy_option='theirs', depth='1', allow_unrelated_histories=True)
+                self.ckm_repo.git_repo.git.gc(prune='all')
                 # Start processing the messages
                 to_delete = []
                 for msg in messages:
--- a/netkan/netkan/repos.py
+++ b/netkan/netkan/repos.py
@@ -39,8 +39,12 @@ class XkanRepo:
         branch = getattr(self.git_repo.heads, branch_name)
         branch.checkout()
 
-    def pull_remote_branch(self, branch_name: str, strategy_option: str = 'ours') -> None:
-        self.git_repo.remotes.origin.pull(branch_name, strategy_option=strategy_option)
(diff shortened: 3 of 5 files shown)


HebaruSan:
```
E             stderr: 'fatal: empty ident name (for <runner@fv-az63-21.hmsqduzlhubuhi2zrxihfknvch.cx.internal.cloudapp.net>) not allowed'
```

Doing some searches, this seems to happen when `user.name` and `user.email` aren't defined in the config. But that should not be needed to pull. :confused:

DasSkelett:
> We're seeing occasional/periodic issues where the CPU credits for the bot plummet sharply and then climb slowly, faster if we force a restart:

Small corrections, for the sake of accuracy:
* That image shows the volume (not CPU) credits, which is what usually is the problem (like the last few times).
* In the two occasions seen in this image it only started to climb again _after_ the restart. The initial fast climbing is with the scheduler skipping runs, and the slower climb is when the scheduler kicks in again >70 credits (seen in the first occasion)
  The second occasion also shows two different rates of plummeting (lol), but both were _before_ the restart. Why and what changed so that the consumption rate decreased at some point is as much of a mystery as this whole bug.

HebaruSan:
> * That image shows the volume (not CPU) credits, which is what usually is the problem (like the last few times).

Fair enough, changed. But CPU is something I understand; what is a "volume credit" tracking?

DasSkelett:
> Fair enough, changed. But CPU is something I understand; what is a "volume credit" tracking?

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html#EBSVolumeTypes_gp2
It's pretty similar to the CPU credits to my understanding, except for the SSD storage that is mounted. According to that text, it allows you to burst beyond the "baseline performance" of the storage.

HebaruSan:
Thanks. That page melted my brain, so I'm going to try to focus on the failing test.

techman83:
I don't think this is how we should address this. All [porcelain operations](http://schacon.github.io/git/git.html#_high_level_commands_porcelain) automatically garbage collect. We already set a git config in the container, we can tune some variables in here to more aggressively gc. For example some operations aren't collected for 90 days, considering our use case, that doesn't make sense and we can tune it to a day etc.

techman83 (changes requested):
I don't think this is how we should address this. All [porcelain operations](http://schacon.github.io/git/git.html#_high_level_commands_porcelain) automatically garbage collect. We already set a git config in the container, we can tune some variables in here to more aggressively gc. For example some operations aren't collected for 90 days, considering our use case, that doesn't make sense and we can tune it to a day etc. It's very likely these changes will do nothing.

I'm not exactly what to tune just yet, but I don't think blindly throwing code here is a good idea. We can wait till the problem arises and then try a couple of different options and see which makes a difference.

HebaruSan:
Fine, but can **somebody** please start an issue where we can collect incident reports for this? I feel like we are starting over every time instead of accumulating data like we ought to.

techman83:
> Fine, but can **somebody** please start an issue where we can collect incident reports for this? I feel like we are starting over every time instead of accumulating data like we ought to.

Yes please.

HebaruSan:
OK, if there are serious reservations about this, then I'm fine with closing it.
But I'm going to pointedly _not_ delete the branch. :smirk:

PR #218, opened 2021-03-29
Fix Scheduler Permissions

techman83:
I got suss when the scheduled task wasn't even showing up in the stopped containers list.

This adds 'CleanCache' to the list of TaskDefinitions that the Event Scheduler is allowed to run.

This fix has been applied and is now working.


HebaruSan (approved):
Nice, that makes sense!

PR #219, opened 2021-05-15
Allow YAML format for .netkans

HebaruSan:
## Motivation

See KSP-CKAN/CKAN#3367, we might want to allow .netkans to use YAML format. If so, the Infra will need to be able to handle that.

## Changes

Now when we load a .netkan file from the repo, we call `yaml.safe_load` instead of `json.loads`, so YAML format is allowed. The return format is the same, and it can still parse JSON files, so there should be no change other than YAML files working.

mypy and friends seem to have gotten smarter and dumber lately, so other updates are made to get the tests passing again.

### DIFF
--- a/netkan/netkan/cli/common.py
+++ b/netkan/netkan/cli/common.py
@@ -14,7 +14,8 @@ from ..github_pr import GitHubPR
 def ctx_callback(ctx: click.Context, param: click.Parameter,
                  value: Union[str, int]) -> Union[str, int]:
     shared = ctx.ensure_object(SharedArgs)
-    setattr(shared, param.name, value)
+    if param.name:
+        setattr(shared, param.name, value)
     return value
 
 
--- a/netkan/.pylintrc
+++ b/netkan/.pylintrc
@@ -11,3 +11,4 @@ disable =
     too-many-return-statements,
     too-many-branches,
     too-many-arguments,
+    consider-using-with,
--- a/netkan/netkan/metadata.py
+++ b/netkan/netkan/metadata.py
@@ -1,4 +1,5 @@
 import json
+import yaml
 import re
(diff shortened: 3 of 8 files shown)


techman83:
Do we have some yaml tests for this?

HebaruSan:
Good point, that definitely needs to be added. :+1:

HebaruSan:
OK, added a test that just parses the YAML .netkan from the PR description and then verifies all the fields. There are probably more complicated things we could check, but that was all that came to mind. Please let me know if you think we should test more.

netkan/netkan/cli/common.py
```diff
@@ -14,7 +14,8 @@
 def ctx_callback(ctx: click.Context, param: click.Parameter,
                  value: Union[str, int]) -> Union[str, int]:
     shared = ctx.ensure_object(SharedArgs)
-    setattr(shared, param.name, value)
+    if param.name:
+        setattr(shared, param.name, value)
```
techman83, line 18: What change required this?

techman83 (approved):
Could be useful for future us to squash up the commits and note in the message what was changed, but I'm not super stressed about it. Go ahead and merge when ready.

netkan/netkan/cli/common.py
HebaruSan: I think that was this:

https://github.com/KSP-CKAN/NetKAN-Infra/runs/2588300928

```
_____________________________ netkan/cli/common.py _____________________________
 note: In function "ctx_callback":
17:21: error: Argument 2 to "setattr" has incompatible
```

Apparently `param.name` isn't guaranteed not to be `None`.

DasSkelett (approved):
Looks good to me as well

PR #222, opened 2021-06-17
Preserve SpaceDock netkan key order

HebaruSan:
## Problem

After #221, KSP-CKAN/NetKAN#8572 was submitted in YAML syntax. It worked, but the properties needed some reordering to match other netkans.

## Cause

PyYAML's `dump` sorts the properties alphabetically by default., which we don't want.

## Changes

Now we pass `sort_keys=False` to keep the original ordering. This will work on Python 3.7 and later with PyYAML 5.1 and later.


HebaruSan:
Self-reviewing so we can see this in action with the next submission...

PR #224, opened 2021-07-02
Increase Instance Memory

techman83:
This is an informative PR as the changes have been applied.

## Problem
With the extra tasks and the growth of the repos, we've started bumping up against the limits of running the service in a single micro.

## Solution
Long term it would be interesting to see if pygit2 was more efficient than gitpython, as it has come a long way since the service was written. In the short term with the testing improvements we did terminate a t2.small, so we have room to go from a t3.micro -> t3.small. I also updated the AMI to the latest available.

Tasks performed (extra to the changes in this pr):
- Created a changeset via aws cli 
  ```
  aws cloudformation create-change-set --stack-name NetKANBot --template-body "`python3 prod-stack.py`" --capabilities 
  CAPABILITY_IAM --profile ckan --region us-west-2 --change-set-name NetkanBot-`date +%s`
  ```
- Instance failed to start the ecs service, shelled in and rebooted it manually so that cloudformation would continue (not sure why)
- Forced a run of the Certbot task to provision the ssl certificate
  ```
   aws ecs run-task --cluster NetKANCluster --task-definition NetKANBotCertBot --profile ckan --region us-west-2
   ```
- Kicked off a inflator schedule for testing
   ```
   aws ecs run-task --cluster NetKANCluster --task-definition NetKANBotScheduler --profile ckan --region us-west-2
   ```
## Results
More memory!
```
              total        used        free      shared  buff/cache   available
Mem:           1954         608          96           0        1249        1204
Swap:             0           0           0
Total:         1954         608          96
```

### DIFF
--- a/netkan/netkan/indexer.py
+++ b/netkan/netkan/indexer.py
@@ -126,11 +126,8 @@ class CkanMessage:
             elif (getattr(status, 'last_warnings', None) != self.WarningMessages and self.WarningMessages is not None):
                 logging.error('New inflation warnings for %s: %s',
                               self.ModIdentifier, self.WarningMessages)
-            actions = [
-                getattr(ModStatus, key).set(
-                    attrs[key]
-                ) for key in attrs
-            ]
+            actions = [getattr(ModStatus, key).set(val)
+                       for key, val in attrs.items()]
             status.update(actions=actions)
         except ModStatus.DoesNotExist:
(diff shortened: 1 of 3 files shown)


HebaruSan (approved):
You get 100 MB, and _you_ get 100 MB, and _you_ ...!

PR #225, opened 2021-07-13
Batch metanetkan download counts

HebaruSan:
## Problem

The bot barfed this up a little while ago:

```
Uncaught exception:
    sys.exit(netkan())
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1137, in __call__
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1668, in invoke
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1404, in invoke
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/decorators.py", line 84, in new_func
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/utilities.py", line 48, in download_counter
    common.token
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/download_counter.py", line 249, in update_counts
    self.get_counts()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/download_counter.py", line 227, in get_counts
    count = netkan_dl.get_count()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/download_counter.py", line 89, in get_count
    count = getattr(self, f'count_from_{self.kref_src}')()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/download_counter.py", line 81, in count_from_netkan
    contents=requests.get(self.remote_netkan).text
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/download_counter.py", line 21, in __init__
    super().__init__(filename, contents)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/metadata.py", line 26, in __init__
    self._raw = yaml.safe_load(self.contents.replace('\t', '    '))
  File "/home/netkan/.local/lib/python3.7/site-packages/yaml/__init__.py", line 162, in safe_load
    return load(stream, SafeLoader)
  File "/home/netkan/.local/lib/python3.7/site-packages/yaml/__init__.py", line 112, in load
    loader = Loader(stream)
  File "/home/netkan/.local/lib/python3.7/site-packages/yaml/loader.py", line 34, in __init__
    Reader.__init__(self, stream)
  File "/home/netkan/.local/lib/python3.7/site-packages/yaml/reader.py", line 74, in __init__
    self.check_printable(stream)
  File "/home/netkan/.local/lib/python3.7/site-packages/yaml/reader.py", line 144, in check_printable
    'unicode', "special characters are not allowed")
yaml.reader.ReaderError: unacceptable character #x0080: special characters are not allowed
  in "<unicode string>", position 10842
```

It suggests that some metanetkan now contains Unicode 0x0080, the PAD character:

- https://www.compart.com/en/unicode/U+0080

I don't know how likely that is to be true, but it's hard to investigate without knowing which mod it is.

## Cause

JSON parsing errors are caught and logged but YAML parsing errors are not.

## Motivation

While investigating the above, I realized (or remembered?) that any metanetkan with a GitHub kref will be handled individually, whereas all other GitHub krefs are processed in batches via GraphQL (see #185). There are ~~at least~~ 68 such mods (~~maybe more if some of the 13 `None`s in my command line output were due to rate limiting without a token~~—these were metanetkans with hard coded `download` properties), which would constitute about 1.4 batches, so this means the download counter currently performs 66 more network requests than strictly needed, with minor excess costs imposed on the container and API rate limiting allowances.

## Changes

- The download counter's error logging is improved:
  - YAML parsing errors are caught and pretty-printed
  - All messages now start with `DownloadCounter`
    - Messages about metanetkans start with `DownloadCounter metanetkan`, to allow us to tell whether a problem happened while processing the main netkan or the metanetkan
  - All messages then print a brief message containing the identifier followed by the source exception
  - One mod with an exception can no longer derail the entire counting loop
- Metanetkans with GitHub krefs are included in the GraphQL batches. This is done via a new `DownloadCounts.resolve_meta()` function that returns the non-metanetkan corresponding to a netkan, either itself or retrieved from the network.

### Debuggability

You can run `python` from the command line and import our classes for interactive investigation. If you evaluate an expression that contains objects, the `__repr__` function is called to generate their string representations. Since we didn't define this for `Netkan` and `Ckan` or their child classes, the default representation of the long type name and memory address is used, which is not useful:

```
[<netkan.download_counter.NetkanDownloads object at 0x7fb10d042610>, <netkan.download_counter.NetkanDownloads object at 0x7fb10d04a490>, <netkan.download_counter.NetkanDownloads object at 0x7fb10d04a130>, ...
```

Now `__repr__` is defined for these classes as `classname(identifier)` or `classname(identifier, version)` depending on which class, so we can tell which mods are being processed. `self.__class__.__name__` is used so inherited classes will also benefit. A similar change is made for `XkanRepo` so we can tell where our repos live.

```
[<NetkanDownloads(AdvancedJetEngine)>, <NetkanDownloads(AlcubierreStandalone)>, <NetkanDownloads(AT-Utils)>, ...

<NetkanRepo(<git.repo.base.Repo '/home/user/Downloads/KSP/NetKAN/.git'>)>
```

### SpaceDock license formatting

We've had a few requests come in from SpaceDock recently with trailing `-` characters in the license. I think people are typing (or pasting) an extra space and we're substituting it. Now we trim excess whitespace from licenses.

- https://github.com/KSP-CKAN/NetKAN/pull/8639/commits/3496514ff821c65dc9fcafa6ca3ceb76ae934f20 (from KSP-CKAN/NetKAN#8639)
- https://github.com/KSP-CKAN/NetKAN/pull/8630/commits/0e516a426290368af48d7bb0dc8dbf87bc2dfa78 (from KSP-CKAN/NetKAN#8630)

### DIFF
--- a/netkan/netkan/download_counter.py
+++ b/netkan/netkan/download_counter.py
@@ -6,7 +6,10 @@ from json.decoder import JSONDecodeError
 from importlib.resources import read_text
 from string import Template
 from typing import Optional, Dict, Any
+
+from yaml.reader import ReaderError
 import requests
+from requests.exceptions import RequestException
 
 from .metadata import Netkan
 from .utils import repo_file_add_or_changed
@@ -41,10 +44,6 @@ class NetkanDownloads(Netkan):
             return f'https://api.cfwidget.com/project/{self.kref_id}'
         return f'https://api.cfwidget.com/kerbal/ksp-mods/{self.kref_id}'
 
-    @property
-    def remote_netkan(self) -> Optional[str]:
-        return self.kref_id
-
     def count_from_spacedock(self) -> int:
         return requests.get(self.spacedock_api).json()['downloads']
 
@@ -75,12 +74,27 @@ class NetkanDownloads(Netkan):
         return requests.get(self.curse_api).json()['downloads']['total']
 
     def count_from_netkan(self) -> int:
-        if self.remote_netkan:
-            return NetkanDownloads(
-                github_token=self.github_token,
-                contents=requests.get(self.remote_netkan).text
-            ).get_count()
-        return 0
+        meta = self.resolve_meta()
+        return meta.get_count() if meta else 0
+
+    def resolve_meta(self) -> Optional['NetkanDownloads']:
+        if self.kref_src == 'netkan' and self.kref_id:
+            try:
+                # Fortunately only one layer of indirection is allowed
+                return NetkanDownloads(
+                    github_token=self.github_token,
+                    contents=requests.get(self.kref_id).text
+                )
+            except RequestException as exc:
+                logging.error('DownloadCounter metanetkan network request failed for %s: %s',
+                              self.identifier, exc)
+            except ReaderError as exc:
+                logging.error('DownloadCounter metanetkan failed YAML parse for %s: %s',
+                              self.identifier, exc)
+            # Can't get the metanetkan
+            return None
+        # Already have the real me
+        return self
 
     def get_count(self) -> int:
         count = 0
@@ -88,20 +102,20 @@ class NetkanDownloads(Netkan):
             try:
                 count = getattr(self, f'count_from_{self.kref_src}')()
             except JSONDecodeError as exc:
-                logging.error(
-                    'Failed decoding count for %s: %s',
-                    self.identifier, exc)
-            except KeyError as exc:
-                logging.error(
-                    'Download count key \'%s\' missing from api for %s',
-                    exc, self.identifier)
-            except requests.exceptions.RequestException as exc:
-                logging.error('Count retrieval failed for %s: %s',
+                logging.error('DownloadCounter failed JSON parse for %s: %s',
                               self.identifier, exc)
-            except AttributeError:
-                # If the kref_src isn't defined, we haven't created a counter
-                # for it and likely doesn't have one.
-                logging.info('Can\'t get count for %s via %s', self.identifier, self.kref_src)
+            except RequestException as exc:
+                logging.error('DownloadCounter network request failed for %s: %s',
+                              self.identifier, exc)
+            except KeyError as exc:
+                logging.error('DownloadCounter key missing for %s: %s',
+                              self.identifier, exc)
+            except AttributeError as exc:
+                logging.error('DownloadCounter attribute missing for %s: %s',
+                              self.identifier, exc)
+            except Exception as exc:  # pylint: disable=broad-except
+                logging.error('DownloadCounter surprising error for %s: %s',
+                              self.identifier, exc)
         return count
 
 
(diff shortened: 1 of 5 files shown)


DasSkelett (approved):
Thanks, looks good. Especially logging with filename/identifier and one failure not halting the entire counter for all mods are long overdue.
Only "negative" impact I can see is that when a source is down (looking at both of you, GitHub and SpaceDock) while the download counter runs, we'll receive a lot of messages, instead of only one. But in those situations we get spammed anyway, so it doesn't make much difference.

DasSkelett:
Right, we probably shouldn't log a missing `count_from_{self.kref_src}` atttribute as error, Jenkins mods for example don't have one:
```
DownloadCounter attribute missing for AmazingCurveEditor:
DownloadCounter attribute missing for CrewManifest:
DownloadCounter attribute missing for CustomBarnKit:
DownloadCounter attribute missing for CustomBarnKit-RO:
DownloadCounter attribute missing for CustomBulkheadProfiles:
DownloadCounter attribute missing for DebugStuff:
DownloadCounter attribute missing for EarlyBird:
DownloadCounter attribute missing for ExtraPlanetaryLaunchpads:
DownloadCounter attribute missing for GCMonitor:
DownloadCounter attribute missing for Graphotron:
DownloadCounter attribute missing for HyperEdit:
DownloadCounter attribute missing for KerbalLaunchVehicles:
DownloadCounter attribute missing for ManeuverD:
DownloadCounter attribute missing for MechJeb2:
DownloadCounter attribute missing for ModularFlightIntegrator:
DownloadCounter attribute missing for ModuleManager:
DownloadCounter attribute missing for SmokeScreen:
DownloadCounter attribute missing for SmokeScreen-RO:
DownloadCounter attribute missing for SpaceDock:
```

What I haven't figured out yet is why the `exc` part is empty.

HebaruSan:
I guess we could define a `count_from_jenkins` function that returns 0 with a TODO.

DasSkelett:
Hm, I'd prefer to keep showing no count for mods where we simply don't have it, instead of an incorrect 0.

HebaruSan:
It's already returning 0 for those, they get excluded:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/8220909ea3c2437c4c685e8651db614d30b3d951/netkan/netkan/download_counter.py#L244-L247

DasSkelett:
Ah, I see.
I also spotted a `#/ckan/http/` mod in the list, there might be some more exotic ones.

HebaruSan:
> we probably shouldn't log a missing `count_from_{self.kref_src}` atttribute as error

I think constructing a string and then using it to call a function is not good coding practice. It's like abusing `eval` in Perl. I would like to refactor this, but I did not think of a good way yet. It seems we are stuck with an `Optional[str]` as our input (because we are starting from `self.kref_src` from the netkans), and a function to call as the output. Maybe just a direct `dict` mapping?

```python
class NetkanDownloads(Netkan):
    # ...
    KREF_HANDLERS = {
        'spacedock': count_from_spacedock,
        'github': count_from_github,
        'curse': count_from_curse,
        'netkan': count_from_netkan,
    }
```

That way we could call `self.KREF_HANDLERS.get` to get the function to call, and handle `None` explicitly rather than using exceptions for flow control. But I'm not sure the class's functions can be referenced during initialization like that, need to test it out.

Or maybe we need to make kref-specific child classes of `NetkanDownloads` (like `GitHubDownloads`, `SpaceDockDownloads`, etc.) that override `get_count`, and then a factory method that generates the right type of object based on the kref. That's probably the right way of representing the kind of behavior that we're trying to achieve here.

techman83:
> Or maybe we need to make kref-specific child classes of `NetkanDownloads` (like `GitHubDownloads`, `SpaceDockDownloads`, etc.) that override `get_count`, and then a factory method that generates the right type of object based on the kref. That's probably the right way of representing the kind of behavior that we're trying to achieve here.

I like this one, I've used a similar pattern recently for a bunch of things.

PR #228, opened 2021-10-13
Batch download counts from SpaceDock

HebaruSan:
## Motivations

See KSP-SpaceDock/SpaceDock#401, getting download counts from SpaceDock currently requires a lot of API calls. That PR is adding a new API route to get them all at once.

Currently the download counter uses the NetKAN repo's .netkan files to find modules. This means we stop counting downloads when a module is frozen.

## Changes

Now the download counter collects all the SpaceDock mods and sends their IDs to the new `/api/download_counts` route, then adds the output to the counts dict.

Note this should not be merged until after KSP-SpaceDock/SpaceDock#401 is merged and live in production. Check the URL with a POST request to confirm:

```
curl -d 'mod_id=127&mod_id=141&mod_id=175' https://spacedock.info/api/download_counts
```

Now the download counter parses the .ckan files in CKAN-meta to find modules for counting, which will work for all modules that you can see in the CKAN client's mod list.

None of the download counter's tests were exercising code that was still in use, so they're removed.


HebaruSan:
Sigh, just got "downvoted" rather than a response. :roll_eyes: 

PyCQA/flake8#1419 reported the issue in more depth, and one of the maintainers just closed it without much explanation. Luckily a later comment suggests `pytest` may be to blame, and a maintainer later said something about "these unsupported hacks". So I guess `pytest` is doing something wrong...?

techman83:
You'll find discussion in pytest-flake8 here -> tholo/pytest-flake8#81

It looks like the author of flake8 isn't interested in supporting the use case of it being wrapped and pytest-flake8 might be using some private flake8 APIs. I saw this in a work project and when I downgraded flake8 it revealed that a double blank line was the issue that tripped it up, so I cleaned that up and decided not to pin at <4.

TL;DR - we can pin to <4 or install 3.9.2 locally and see if there is a test failing that is being hidden by the failure.

HebaruSan:
They're rapidly locking down all the issues where I've tried to get anything other than a dismissive non-answer, most recently PyCQA/flake8#1445.

Thanks, I'll try that pin in a moment...

HebaruSan:
The middle of that discussion makes it sound like they don't want people to use flake8 this way, and it officially ["does not have a public, stable Python API"](https://flake8.pycqa.org/en/latest/user/python-api.html#public-python-api). With community attempts to fix the problem being rejected, continuing to try to use flake8 seems like swimming against the tide. Removed it in the latest commit.

techman83 (approved):
LGTM, should make this process faster!

HebaruSan:
KSP-SpaceDock/SpaceDock#401 finally merged with KSP-SpaceDock/SpaceDock#452 and should be live now. Let's try this!

HebaruSan:
Download counts are now completely broken. :roll_eyes: 

https://github.com/KSP-CKAN/CKAN-meta/commit/71fdc60809676aba8aa02f7261ab2dd3f0afef70

image

PR #230, opened 2021-10-22
Standardize branch names, set pull request labels

HebaruSan:
## Motivation

- Branch names generated by the bot are inconsistent
  - Auto Freezer: `freeze/auto`
  - SpaceDock Adder: `add-identifier`
  - Indexer: `identifier-version`
- Pull requests submitted by the bot currently have no labels
- It was inconvenient that the Auto Freezer's commits were in alphabetical order while its pull request body listed the mods in chronological order

## Changes

- Now the branch names are consistent
  - Auto Freezer: `freeze/auto`
  - SpaceDock Adder: `add/identifier`
  - Indexer: `add/identifier-version`
- Now pull requests have labels
  - Auto Freezer: `Pull request`, `Freeze`, `Needs looking into`
  - SpaceDock Adder: `Pull request`, `Mod request`
  - Indexer: `Needs looking into`
- Now the Auto Freezer's commits and pull request body are both in chronological order

PR #231, opened 2021-11-13
Set pull request labels in separate API request

HebaruSan:
## Problem

The label part of #230 hasn't worked; pull requests submitted subsequently have had no labels.

## Cause

The pull request API accepts a fragment of a pull request object including the title, body, etc., but the `labels` property is missing:

- https://docs.github.com/en/rest/reference/pulls#create-a-pull-request

But at a database level, a pull request is implemented as a special type of issue, and labels can be set at issue creation:

- https://docs.github.com/en/rest/reference/issues#create-an-issue

Surely this is also possible for pull requests, and they just forgot to include it in the documentation?

No:  https://github.community/t/create-pull-request-and-add-labels-in-a-single-request/116146

## Changes

Now if you pass labels to a PR creation call, we make a second API call to set them.

I am planning to self-review and merge this so I can see whether it works the next time we have a PR from the bot.

### DIFF
--- a/netkan/netkan/github_pr.py
+++ b/netkan/netkan/github_pr.py
@@ -25,7 +25,6 @@ class GitHubPR:
                 'base': 'master',
                 'head': branch,
                 'body': body,
-                'labels': labels or [],
             }),
         )
         if response.status_code not in [200, 201, 204]:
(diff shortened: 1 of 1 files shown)


HebaruSan:
Worked! KSP-CKAN/NetKAN#8863:

image

PR #234, opened 2021-12-03
Filter out clutter files in SD PRs

HebaruSan:
## Motivation

After #233, I remembered that @linuxgurugamer had also submitted KSP-CKAN/NetKAN#5178, so I decided to mine it for ideas. It mentions automatic `filter` and `filter_regexp` property generation based on ZIP contents and special handling for `Ships` folders.

## Changes

Now the `ModAnalyzer` scans the ZIP to see if any of its files match several common filters, which I determined by looking at all the filters currently present in the NetKAN repo. If so, the install stanza is modified to filter them out. This should take another small bit of work off the hands of maintainers.

`Ships` folders are **not** handled specially, because there's too much variation in how authors treat those folders to handle them automatically, and we have an inflator warning for that already anyway (see KSP-CKAN/CKAN#3207).

I'm planning to self-review this since it's mainly for me anyway.

### DIFF
--- a/netkan/netkan/mod_analyzer.py
+++ b/netkan/netkan/mod_analyzer.py
@@ -2,7 +2,7 @@ import re
 import tempfile
 from pathlib import Path
 from zipfile import ZipFile, is_zipfile
-from typing import Dict, Any
+from typing import Dict, List, Any, Union
 import requests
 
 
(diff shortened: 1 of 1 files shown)


techman83:
Neat!

PR #243, opened 2021-12-10
Stream downloads

HebaruSan:
## Problem

See #242, the `ModAnalyzer` can throw `OSError: [Errno 14] Bad address`.

## Cause

Apparently that exception means "out of memory". The `SpaceDockAdder` instance is configured to have 128 MiB of RAM, and it was trying to download a 250 MiB file.

I just wanted to dump the response into the file, but `Response.content` has to download the entire file into memory first.

## Changes

Now the downloading code is refactored into `common.download_stream_to_file`, which retrieves and stores one 8 KiB chunk at a time. This also allowed the consolidation of the `User-Agent` handling. After this the `ModAnalyzer` and the `Mirrorer` should be able to handle big files using less memory.

### DIFF
--- a/netkan/netkan/common.py
+++ b/netkan/netkan/common.py
@@ -1,11 +1,14 @@
-from typing import List, Iterable, Dict, Any
+from typing import List, Iterable, Dict, Any, IO
 import boto3  # pylint: disable=unused-import
 
+import requests
 import github
 from git import Repo
 
 from .metadata import Netkan
 from .repos import NetkanRepo
+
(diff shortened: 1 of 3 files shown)


DasSkelett (approved):
> This also allowed the consolidation of the User-Agent handling. 

That's nice!
Let's give it a try. Just have to hope a bigger-ish mod gets uploaded to SpaceDock again soon

techman83:
Awesome, looks like a neat solution.

PR #252, opened 2022-01-05
Check archive.org overloading before each Mirrorer upload

HebaruSan:
## Problem

```
Uncaught exception:
Traceback (most recent call last):
  File "/home/netkan/.local/lib/python3.7/site-packages/internetarchive/item.py", line 1091, in upload_file
    response.raise_for_status()
  File "/home/netkan/.local/lib/python3.7/site-packages/requests/models.py", line 953, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 503 Server Error: Slow Down for url: https://s3.us.archive.org/ROEngines-v1.10.0.0/A196AD37-ROEngines-v1.10.0.0.zip

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".local/bin/netkan", line 8, in <module>
    sys.exit(netkan())
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/services.py", line 76, in mirrorer
    ).process_queue(common.queue, common.timeout)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/mirrorer.py", line 318, in process_queue
    if self.try_mirror(CkanMirror(self.ia_collection, path)):
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/mirrorer.py", line 341, in try_mirror
    ckan.download_headers)
[...]

PR #253, opened 2022-01-05
Use API call to check if archive.org overloaded

HebaruSan:
## Background

The Mirrorer's current check for whether archive.org is overloaded is peformed by creating a URL string, submitting it via `requests.get`, and parsing the results (based on how the old Perl code used to do the same thing). This embeds extra knowledge about archive.org's operations in the Mirrorer.

## Motivation

After #252 I wondered whether there was a more standard way to do this. And sure enough, there is a function built into the `internetarchive` module for it!

https://github.com/jjjake/internetarchive/blob/a4d5cfdafead333ed18fa78e78f146a6bf670fa9/internetarchive/session.py#L340-L362

Differences:

- That function uses an optional `bucket` parameter that our code doesn't use
- Our code checks for HTTP status 503 and theirs doesn't

I think these differences are unlikely to change anything, and even if they do, the module's approach is probably better than ours since it will be used by other people.

## Changes

Now we call `ArchiveSession.s3_is_overloaded` instead of rolling our own HTTPS request.

### DIFF
--- a/netkan/netkan/mirrorer.py
+++ b/netkan/netkan/mirrorer.py
@@ -9,7 +9,6 @@ from pathlib import Path
 from importlib.resources import read_text
 from typing import Optional, List, Union, Iterable, BinaryIO, Dict, Any
 import boto3
-import requests
 import github
 import internetarchive
 from jinja2 import Template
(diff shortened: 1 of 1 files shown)


techman83:
Good catch! Yes when the perl code was written, it was pretty early in the phase of this functionality.

HebaruSan:
> Good catch! Yes when the perl code was written, it was pretty early in the phase of this functionality.

To be fair, the Perl code also didn't have the advantage of using a third party library specifically intended for interfacing with archive.org; it was trying to _be_ that library itself. Now we can stand on the shoulders of jjjake.

PR #260, opened 2022-01-29
Submit pull requests via Github library

HebaruSan:
## Motivation

The Indexer, AutoFreezer, and SpaceDockAdder use the `GitHubPR` class to submit pull requests. Internally, this class uses the `requests` and `json` libraries to create and submit HTTPS messages to hard coded GitHub API URLs, then parse the responses.

There's a `github` library that abstracts this logic for us, which we are already using in a few other places.

## Changes

`GitHubPR` now uses the `github` library and is much, much shorter.

All of the other places that were already using the `github` library now set the user agent from #236.

This one should probably be reviewed, as there may be functional changes, especially since the error handling has been changed from parsing JSON to catching exceptions thrown by the library.

### DIFF
--- a/netkan/netkan/common.py
+++ b/netkan/netkan/common.py
@@ -34,7 +34,7 @@ def pull_all(repos: Iterable[Repo]) -> None:
 
 
 def github_limit_remaining(token: str) -> int:
-    return github.Github(token).get_rate_limit().core.remaining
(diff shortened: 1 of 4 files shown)


techman83 (approved):
Yep, we already use the dependency elsewhere, so it makes sense to use it here. LGTM.

PR #266, opened 2022-04-20
Delete cached downloads when a mod is frozen

HebaruSan:
## Problem

image

## Cause

The disk may be full, but I'm skeptical of that because this download is only 278 MB:

https://github.com/SpaceODY/PRVE/releases/tag/v1.7.1

Regardless, I had an idea for saving space.

## Changes

Now when we freeze a mod, we remove its downloads from the bot's cache, since they won't be needed for future bot passes. This should help to reduce the risk of the disk filling up. If we need to unfreeze the mod in the future, then it would be re-downloaded at that time.

This will only affect freshly frozen mods; we will not attempt to purge mods that were frozen in the past.


DasSkelett (approved):
Sounds good

PR #268, opened 2022-05-05
Update ami + increase cache disk

techman83:
These changes have already been applied, but updating the configuration for transparency

### DIFF
--- a/prod-stack.py
+++ b/prod-stack.py
@@ -628,7 +628,7 @@ docker_service = InitService(
 netkan_instance = Instance(
     'NetKANCompute',
     # ECS Optimised us-west-2
-    ImageId='ami-064803387adcb64b3',
+    ImageId='ami-0e0e34cdb5fd714fc',
     InstanceType='t3.small',
     IamInstanceProfile=Ref(netkan_profile),
     KeyName='techman83_alucard',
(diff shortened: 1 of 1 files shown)


HebaruSan (approved):
I didn't know what this did, but apparently we're increasing from 50 GiB to 100 GiB:

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_EbsBlockDevice.html

> VolumeSize (request), volumeSize (response)
> The size of the volume, in GiBs. You must specify either a snapshot ID or a volume size. If you specify a snapshot, the default is the snapshot size. You can specify a volume size that is equal to or larger than the snapshot size.
> 
> The following are the supported volumes sizes for each volume type:
> 
> - gp2 and gp3:1-16,384
> - io1 and io2: 4-16,384
> - st1 and sc1: 125-16,384
> - standard: 1-1,024
> 
> Type: Integer
> 
> Required: No

techman83:
Ahh, sorry for the lack of explanation. Seems things like this only crop up when I'm traveling! And yes, the cache disk has been increased from 50 -> 100.

HebaruSan:
> Ahh, sorry for the lack of explanation.

No worries mate, it gave me a chance to learn more about AWS. :+1:

PR #275, opened 2022-09-06
Get download counts from archive.org

HebaruSan:
## Motivation

Currently the download counter only works with GitHub and SpaceDock. Archive.org has a view count API, but we don't use it:

https://archive.org/services/docs/api/views_api.html

(As far as I can tell, this isn't supported by the `internetarchive` Python module.)

There are several mods with a primary download URL on archive.org in CKAN, and a few with over 1000 downloads:

```
$ sort -rn -k 2 download_counts.json | head -n 40
    "MiniAirbrakes": 26556,
    "ProceduralParts-Textures-SCCKSCS": 15520,
    "FederalProductions-Flags": 8613,
    "AlternisKerbolRekerjiggered-flags": 7928,
    "B9AerospaceProceduralParts": 2844,
    "NearFuturePropulsionExtras": 1781,
    "KWRocketry": 1719,
    "SpaceXLandingLegs": 1533,
    "FShangarExtender": 1341,
    "BDArmoryFPS": 1303,
    "KaribouExpeditionRover": 1073,
    "DoubleDouble": 1063,
    "ContractConfigurator-AdvancedProgression": 1057,
    "kOS-Classic": 1003,
```

## Changes

- Now a new `InternetArchiveBatchedQuery` class works like the classes for GitHub and SpaceDock, but with the IA API
- Now `DownloadCounter` creates an `InternetArchiveBatchedQuery` object and feeds it mods with download URLs on archive.org, submitting requests for each group of 30

### Known limitations

Archive.org's view stats are per-IA-identifier, which for CKAN's archived mods means per-mod-version. To get the stats for all versions we would have to loop over all of them.

For now we'll only look at the most recent version of each mod. This should be fine; since we only move mods to a primary archive.org download if they are abandoned and the original hosting becomes unavailable, there won't be many downloads on the older archived versions.

### DIFF
--- a/netkan/netkan/download_counter.py
+++ b/netkan/netkan/download_counter.py
@@ -11,6 +11,7 @@ import requests
 
 from .utils import repo_file_add_or_changed
 from .repos import CkanMetaRepo
+from .metadata import Ckan
 
 
 class GraphQLQuery:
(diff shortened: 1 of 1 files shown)


techman83 (approved):
That seems neat enough. I wonder if downloads count against this though?

HebaruSan:
> That seems neat enough. I wonder if downloads count against this though?

Yes, I think "views" does mean "downloads", since archive.org was originally about hosting documents for reading. A page with 0 views stays that way even if I load and reload the summary page:

https://archive.org/details/Shaddy-v2.5

HebaruSan:
Regarding KSP-CKAN/CKAN#3661 and the new GitLab krefs, [GitLab's documentation]() says that its download stats are only available to members of the project in question, and it sounds like it only counts `git clone` operations, so I won't be trying to access them:

https://docs.gitlab.com/ee/api/project_statistics

> Retrieving these statistics requires write access to the repository.

PR #280, opened 2023-02-18
Check scheduler limits before cloning repos

HebaruSan:
## Problem

During periods of low system resource availability, the Scheduler can fail before it even checks whether it's supposed to run:

```
Uncaught exception:
Traceback (most recent call last):
  File ".local/bin/netkan", line 8, in <module>
    sys.exit(netkan())
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/services.py", line 60, in scheduler
    common.netkan_repo, common.ckanmeta_repo, common.queue, common.token,
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/common.py", line 68, in __getattribute__
    attr = super().__getattribute__(name)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/common.py", line 101, in ckanmeta_repo
    init_repo(self._ckanmeta_remote, '/tmp/CKAN-meta', self.deep_clone))
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/utils.py", line 14, in init_repo
    Repo.clone_from(metadata, clone_path, depth=1))
  File "/home/netkan/.local/lib/python3.7/site-packages/git/repo/base.py", line 1275, in clone_from
    return cls._clone(git, url, to_path, GitCmdObjectDB, progress, multi_options, **kwargs)
[...]

PR #281, opened 2023-02-20
Better AutoFreezer PR continuation

HebaruSan:
## Problem

I let KSP-CKAN/NetKAN#9471 sit without review for quite a while, and in the meantime we had 9 weekly checks go by without any new commits pushed. Probably quite a few abandoned mods have slipped through the cracks, which we should probably catch up at some point.

## Cause

The branch checkout logic in the AutoFreezer was trying to re-use an existing branch if there was one, but it was missing a step: it fetched the remote branch, and it tried to check it out, but it never created a local branch corresponding to the remote branch. So checking it out failed, and it always fell back to creating a new branch. Then later the attempt to push to the server would fail because the branch already existed with conflicts.

## Changes

- Now the AutoFreezer uses the shared `NetkanRepo.change_branch` function instead of rolling its own version of that logic, which will fix the problem and reduce code duplication
- To help with review, the bot will now post its intended PR body as a comment when it pushes commits to an existing open PR. This way the person reviewing will see the table of the newly added mods with their last update and links.

This way if we miss a week of review, the next week's idle mods will get cleanly added on to the existing PR, and none will be missed.

Note that this branch shares the test-fixes commit with #280, but not the one with that PR's actual functionality. I think we can merge either one first or not at all and the other one should be fine to continue or not.

### DIFF
--- a/netkan/netkan/auto_freezer.py
+++ b/netkan/netkan/auto_freezer.py
@@ -1,7 +1,6 @@
 import logging
 from datetime import datetime, timezone, timedelta
 from typing import Iterable, Optional, List, Tuple
-import git
 
 from .status import ModStatus
 from .repos import NetkanRepo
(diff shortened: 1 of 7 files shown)


techman83 (approved):
That all makes sense, lgtm!

HebaruSan:
Huh, that just didn't work at all.

KSP-CKAN/NetKAN#9580

image

Not even an exception to investigate.

PR #283, opened 2023-03-04
Refactor Game Management

techman83:
This is a refactor of the game management to introduce Multi Game support. Whilst there are a large number of lines added, about a 1/3 are test cases and of the code changed, there is now 79% coverage and increased overall coverage from 54% to 67%.

The configuration uses the multi var Click option to achieve the multi game config, which is done via `--option var --option var2` or a space separated env var `OPTION=var var`. This allows a configuration of `game=variable`, ie `ksp=CKANmeta`

## Out of Scope for this PR
It was already looking like a large set of changes, so some decisions were made to at least constrain it somewhat. Allowing to at least ensure the status quo is working before iterating to the next set of changes

- Mirrorer - The intention would be to use the generic queue implementation, bringing us to a single queue implementation. However we can't apply for a collection until we have 50 mods to upload.
- Primary branch from `master` to `main` - this seemed sensible to refactor independently
- Scheduler/Freezer etc to iterate through available games
- Common/SharedArgs naming convention - this could use improvement. Beyond the entry point, the rest of the application probably doesn't need to care about the arguments, just that they are available. With some more test coverage this could be refactored separately to this PR
- ModStatus + how we handle identifying/separating of the games

## Added
- Coverage Reporting
- `Game` class, that returns game specific values. There is scope here for improvement, but the current implementation does work and will benefit from being used
- game-id option - this allows the existing utilities to operate as is without much change
- inflation-queues option - This is a multi var to direct inflation to the relevant queues
- `BaseMessageHandler` to provide a consistent interface for handling SQS messages
- `QueueHandler` move queue logic out of cli, and into a testable common class

## Changed
- ckanmeta-remote -> ckanmeta-remotes - multi game repos ckanmeta
- netkan-remote -> netkan-remotes - multi game repos for netkan
- ia-collection -> ia-collections - allow for the configuration of multi game internet archive repos
- `GitHubPR` - repo changed to lazy load, avoiding it trying to auth unnecessarily on class instantiation.
- `SpaceDockAdder` moved to common queue/message handler
- Webhooks multi-game handling
- Webhooks queues changed to lazy load - speeds up service start and simplifies testing

## Improved
- SharedArgs variables now typed at a class level and not set during init. Improves type hints during coding, and reduces likelihood of a null where it's not expected
- Boto3 type checking via mypy-boto3 - caveat of it needing `if TYPE_CHECKING:` to avoid upsetting pylint

*Note*: Merging will need to coincide with deployment of the updated configuration. I will allocate time to do this, and monitor/test relevant services after.


HebaruSan:
SpaceDock is just about ready now (see KSP-SpaceDock/SpaceDock#480). Is there anything still missing from this, or could we merge it and fix things as they break?

netkan/netkan/cli/common.py
```diff
@@ -48,27 +54,140 @@ def ctx_callback(ctx: click.Context, param: click.Parameter,
...
+    name: str
+    shared: 'SharedArgs'
+    clone_base: str = '/tmp'
+    _ckanmeta_repo: CkanMetaRepo
+    _ckanmeta_remote: str
+    _netkan_repo: NetkanRepo
+    _netkan_remote: str
+    _github_pr: GitHubPR
+    _ia_collection: str
+    _infaltion_queue: str
```
HebaruSan, line 75:
```suggestion
    _inflation_queue: str
```

netkan/netkan/cli/common.py
```diff
@@ -48,27 +54,140 @@ def ctx_callback(ctx: click.Context, param: click.Parameter,
...
+    def ia_collection(self) -> str:
+        if getattr(self, '_ia_collection', None) is None:
+            self._ia_collection = self.args('ia_collections')
+        return self._ia_collection
+
+    @property
+    def inflation_queue(self) -> str:
+        if getattr(self, '_inflation_queue', None) is None:
+            self._infaltion_queue = self.args('inflation_queues')
+        return self._infaltion_queue
```
HebaruSan, line 158:
```suggestion
            self._inflation_queue = self.args('inflation_queues')
        return self._inflation_queue
```

netkan/netkan/common.py
```diff
@@ -52,3 +70,89 @@ def download_stream_to_file(download_url: str, dest_file: IO[bytes]) -> None:
...
+        self.game = game
+
+    # Apparently gitpython can be leaky on long running processes
+    # we can ensure we call close on it and run our handler inside
+    # a context manager
+    def __enter__(self) -> 'BaseMessageHandler':
+        if not self.repo.is_active_branch('master'):
+            self.repo.checkout_branch('master')
+        self.repo.pull_remote_branch(
+            'master', strategy_option=self.STRATEGY_OPTION)
```
HebaruSan, line 89: Can we do better than hard-coding `master` in new code now that we know we need to change it?

netkan/netkan/indexer.py
```diff
@@ -154,45 +163,54 @@ def process_ckan(self) -> None:
...
+        if getattr(self, '_staged', None) is None:
+            self._staged = deque()
+        return self._staged
+
+    @property
+    def processed(self) -> List[CkanMessage]:
+        if getattr(self, '_processed', None) is None:
+            self._processed = []
+        return self._processed
+
```
HebaruSan, line 209: Are a `deque` or a list really large enough to be worth the added complexity of lazy initialization? This looks like textbook premature optimization.

netkan/netkan/webhooks/__init__.py
```diff
@@ -33,12 +33,12 @@ def __init__(self) -> None:
...
-        os.environ.get('INFLATION_SQS_QUEUE', ''),
-        os.environ.get('ADD_SQS_QUEUE', ''),
-        os.environ.get('MIRROR_SQS_QUEUE', '')
+        ssh_key=os.environ.get('SSH_KEY', ''),
+        secret=os.environ.get('XKAN_GHSECRET', ''),
+        netkan_remote=os.environ.get('NETKAN_REMOTE', ''),
+        ckanmeta_remote=os.environ.get('CKANMETA_REMOTE', ''),
+        inf_queue_name=os.environ.get('INFLATION_SQS_QUEUE', ''),
+        add_queue_name=os.environ.get('ADD_SQS_QUEUE', ''),
+        mir_queue_name=os.environ.get('MIRROR_SQS_QUEUE', '')
```
HebaruSan, line 42: Aren't the webhooks going to need the capability to handle multiple repos and queues? It's weird to see the single-repo assumption carried forward here in the one container that definitely needs to remove it.

netkan/netkan/cli/common.py
techman83: Hahahaha! Goodness, what a typo! Excellent spot :slightly_smiling_face:

netkan/netkan/common.py
```diff
@@ -1,24 +1,42 @@
...
+        DeleteMessageBatchRequestEntryTypeDef,
+        SendMessageBatchRequestEntryTypeDef,
+    )
+    from .cli.common import Game, SharedArgs
+else:
+    Game = object
+    Message = object
+    DeleteMessageBatchRequestEntryTypeDef = object
+    SendMessageBatchRequestEntryTypeDef = object
+    SharedArgs = object
```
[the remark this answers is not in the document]
techman83: This just allows us to accurately type check it with dynamic stuff that comes with boto3. This gives us type checking without breaking the runtime code. See -> https://mypy-boto3.readthedocs.io/en/latest/#pylint-compatibility

netkan/netkan/cli/utilities.py
```diff
@@ -30,8 +30,8 @@
 @pass_state
 def auto_freezer(common: SharedArgs, days_limit: int, days_till_ignore: int) -> None:
     afr = AutoFreezer(
-        common.netkan_repo,
-        common.github_pr,
+        common.game(common.game_id).netkan_repo,
+        common.game(common.game_id).github_pr,
```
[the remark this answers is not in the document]
techman83: Yeah, I'm open to that. But wanted to leave how we handled that to a follow up PR. This one is already quite weighty.

netkan/netkan/common.py
techman83: Yes. I want to refactor our git usage in an upcoming PR.

netkan/netkan/common.py
```diff
@@ -52,3 +70,89 @@ def download_stream_to_file(download_url: str, dest_file: IO[bytes]) -> None:
...
+        raise NotImplementedError
+
+
+class QueueHandler:
+    common: SharedArgs
+    _game_handlers: Dict[str, BaseMessageHandler]
+    _handler_class: Type[BaseMessageHandler] = BaseMessageHandler
+
+    def __init__(self, common: SharedArgs) -> None:
+        self.common = common
```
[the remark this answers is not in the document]
techman83: Open to refactoring it in a future PR. A lot has changed since the initial codebase was formed. We definitely can't use `args` though :slightly_smiling_face:

netkan/netkan/indexer.py
```diff
@@ -154,45 +163,54 @@ def process_ckan(self) -> None:
...
-class MessageHandler:
-
-    def __init__(self, repo: CkanMetaRepo, github_pr: Optional[GitHubPR] = None) -> None:
-        self.ckm_repo = repo
-        self.github_pr = github_pr
-        self.master: Deque[CkanMessage] = deque()
-        self.staged: Deque[CkanMessage] = deque()
-        self.processed: List[CkanMessage] = []
+class MessageHandler(BaseMessageHandler):
+    _master: Deque[CkanMessage]
```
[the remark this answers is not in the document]
techman83: Yeah, I'll look at that when I refactor the queue handling. I'm thinking `_primary`

netkan/netkan/indexer.py
techman83: It's always been a deque. But it's not so much for optimisation, rather convenience. Allows the queue to be processed in the order it came in via `popleft`. It's worked for nearly 4 years, so I think I'm happy to keep it the way it is.

netkan/netkan/webhooks/__init__.py
techman83: Yeah I just noticed that, I will update the names to be less confusing

techman83:
> * This is a LOT of new code, and new code means new bugs (and a lot of review time)

The bulk of the lines added are test cases. The bulk of the new code comes with test coverage increases.

> * The existing code has been made a LOT more complex (increased complexity also means new bugs as well as harder-to-fix bugs)

I'm not sure I agree entirely here. Admittedly the multi-game was a bit more work than I anticipated. However the queue handling in particular has been generalised, it's a pretty generic task and we had three different implementations. We're down to 2, but I left the mirrorer as a separate exercise as we can't add it for ksp2 just yet and I figured the PR was big enough as it is.

> * These containers simply don't need to and shouldn't know what game they're operating on; the only behavior that it determines is which repos and queues to use, which was already well handled by the previously existing structure. Everything else they do is the exact same logic regardless of what game it is, so passing around `Game` objects and `game_id` values is pure cost in terms of technical debt and cognitive load for no benefit (asterisk for webhooks).

I think there is scope to improve the game handling. Especially for the generalised services that do things like scheduling.

> * It's been almost three weeks and this PR isn't even out of draft status yet (not a criticism of you, but it is a reflection of the contrast between the scope of this design and our available resources)

It is unfortunate that it coincided with a Conference + Holiday, but I hear what you are saying. Reading through the code there are a bunch of opportunities to simplify and refactor. But I left them out of scope as this PR was big enough.

> I'm having second thoughts about closing https://github.com/KSP-CKAN/NetKAN-Infra/pull/282 now that I've seen the details of this code I'm potentially going to have to co-maintain going forward. We probably could have wrapped that one up in another day or two of dev on the webhooks and gone live with KSP2 support weeks ago. Is there anything I can say to persuade you to consider the approach that leverages proven, stable code?

Yes, I'm pretty committed to this. Better multi-game handling was always the goal, and simply duplicating the infrastructure isn't super scalable and didn't address the webhooks.

The two things that will need to happen next
- Refactor repo primary branch handling
- Enable KSP2 config
- Refactor mirroring

techman83:
I've gotta head to a flight, but happy to update the PR info and maybe take you on a tour of the changes @HebaruSan 

I have done a bunch of e2e testing, and I'm confident the actual logic hasn't changed, more the entry points.

The Merge will need to happen alongside the prod deployment changes, and I'll be happy to dedicate time to doing that to make sure it all goes smoothly/works.

netkan/netkan/indexer.py
HebaruSan: Oh, I'm absolutely fine with using a `deque`. My question is whether these properties are needed, rather than just doing this in `__init__`:

```python
    self.master = deque()
    self.staged = deque()
    self.processed = []
```

What do we get from making these properties lazily initialized instead?

netkan/netkan/common.py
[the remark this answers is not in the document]
HebaruSan: But this is all new code. The `Game` class is completely new. Getting the name right now makes more sense to me than creating a poor name and deferring a good name to the indefinite future.

netkan/netkan/common.py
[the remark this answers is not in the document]
HebaruSan: What's the benefit over importing the types unconditionally? I.e., just doing this at the top like normal:

```python
from mypy_boto3_sqs.service_resource import Message
from mypy_boto3_sqs.type_defs import (
    DeleteMessageBatchRequestEntryTypeDef,
    SendMessageBatchRequestEntryTypeDef,
)
from .cli.common import Game, SharedArgs
```

netkan/netkan/common.py
```diff
@@ -52,3 +70,89 @@ def download_stream_to_file(download_url: str, dest_file: IO[bytes]) -> None:
                       timeout=60) as req:
         for chunk in req.iter_content(chunk_size=8192):
             dest_file.write(chunk)
+
+
+class BaseMessageHandler:
```
HebaruSan, line 75: Let's put `BaseMessageHandler` and `QueueHandler` in their own file(s). They look big and complex enough to deserve that; `common.py` is a surprising place to find them.

HebaruSan:
> However the queue handling [...] had three different implementations.

Yeah, that was very much intentional. Retrieving messages from a queue and deleting them is, frankly, too simple of a task to be worth centralizing. (To me it's a bit like making a wrapper around 
`logging.error` to reduce the number of "implementations" of logging an error message; doing the thing itself instead is simpler and more maintainable.) Especially when it means that key logic that I need to read to understand what it's doing and investigate/fix bugs is now spread across multiple different files rather than in a simple "X then Y then Z" procedural code block.

> We're down to 2

... so naturally I'm not delighted about this. ☝️

netkan/netkan/common.py
```diff
@@ -52,3 +70,89 @@ def download_stream_to_file(download_url: str, dest_file: IO[bytes]) -> None:
...
+            )
+            if not messages:
+                continue
+            for message in messages:
+                game = message.message_attributes.get(  # type: ignore[union-attr,call-overload]
+                    'GameId', {}).get('StringValue', None)
+                if game is None:
+                    logging.error('GameId missing from MessageAttributes')
+                    continue
+                self.append_message(game, message)
```
HebaruSan, line 151:
```suggestion
    def game_handler(self, game_id: str) -> BaseMessageHandler:
        if self.game_handlers.get(game_id, None) is None:
            self.game_handlers.update({
                game: self._handler_class(self.common.game(game_id))
            })
        return self.game_handlers[game_id]

    def append_message(self, game_id: str, message: Message) -> None:
        self.game_handler(game_id).append(message)

    def run(self) -> None:
        sqs = boto3.resource('sqs')
        queue = sqs.get_queue_by_name(QueueName=self.common.queue)
        while True:
            messages = queue.receive_messages(
                MaxNumberOfMessages=10,
                MessageAttributeNames=['All'],
                VisibilityTimeout=self.common.timeout
            )
            if not messages:
                continue
            for message in messages:
                game_id = message.message_attributes.get(  # type: ignore[union-attr,call-overload]
                    'GameId', {}).get('StringValue', None)
                if game_id is None:
                    logging.error('GameId missing from MessageAttributes')
                    continue
                self.append_message(game_id, message)
```

Let's try to use `game_id` consistently when the variable is a string.

netkan/netkan/scheduler.py
```diff
@@ -30,13 +38,13 @@ def __init__(self, common: SharedArgs, queue: str, github_token: str,
 
     @property
     def nk_repo(self) -> NetkanRepo:
-        return self.common.netkan_repo
+        return self.common.game(self.game).netkan_repo
```
HebaruSan, line 41:
```suggestion
        return self.common.game(self.game_id).netkan_repo
```

netkan/netkan/scheduler.py
```diff
@@ -30,13 +38,13 @@ def __init__(self, common: SharedArgs, queue: str, github_token: str,
...
 
     @property
     def ckm_repo(self) -> CkanMetaRepo:
-        return self.common.ckanmeta_repo
+        return self.common.game(self.game).ckanmeta_repo
```
HebaruSan, line 45:
```suggestion
        return self.common.game(self.game_id).ckanmeta_repo
```

netkan/netkan/common.py
```diff
@@ -52,3 +70,89 @@ def download_stream_to_file(download_url: str, dest_file: IO[bytes]) -> None:
...
+        if getattr(self, '_game_handlers', None) is None:
+            self._game_handlers = {}
+        return self._game_handlers
+
+    def game_handler(self, game: str) -> BaseMessageHandler:
+        if self.game_handlers.get(game, None) is None:
+            self.game_handlers.update({
+                game: self._handler_class(self.common.game(game))
+            })
+        return self.game_handlers[game]
```
HebaruSan, line 129:
```suggestion
    def game_handler(self, game_id: str) -> BaseMessageHandler:
        if self.game_handlers.get(game_id, None) is None:
            self.game_handlers.update({
                game_id: self._handler_class(self.common.game(game_id))
            })
        return self.game_handlers[game_id]
```

netkan/netkan/common.py
```diff
@@ -52,3 +70,89 @@ def download_stream_to_file(download_url: str, dest_file: IO[bytes]) -> None:
...
+
+    def append_message(self, game: str, message: Message) -> None:
+        self.game_handler(game).append(message)
```
HebaruSan, line 132:
```suggestion
    def append_message(self, game_id: str, message: Message) -> None:
        self.game_handler(game_id).append(message)
```

netkan/netkan/common.py
HebaruSan, line 151:
```suggestion
                game_id = message.message_attributes.get(  # type: ignore[union-attr,call-overload]
                    'GameId', {}).get('StringValue', None)
                if game_id is None:
                    logging.error('GameId missing from MessageAttributes')
                    continue
                self.append_message(game_id, message)
```

netkan/netkan/cli/utilities.py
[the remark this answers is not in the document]
HebaruSan: > a follow up PR. This one is already quite weighty.

Frankly I'd rather have everything in one big bundle, so I don't have to guess what the end state will look like while reviewing some portion of the changes or guess what will and will not be addressed elsewhere. Can we merge all these PRs together?

HebaruSan (changes requested):
Added some replies to previous comments, then tried to figure out how the SpaceDockAdder works now, which led me down the rabbit hole of "when is `game` getting set," which was confusing because `game` might be a `str` or a `Game` depending on context. Added suggestions for those.

netkan/netkan/indexer.py
techman83: It's less work to do within the init, and means we don't need to overwrite it during inheritance. It's wasn't so much for performance reasons, just a combination of my preference for lazy loading and a slightly cleaner implementation when it comes to inheritance.

netkan/netkan/common.py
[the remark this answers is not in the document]
techman83: The future won't be indefinite, and I agree with you. I just would rather get this changeset out and in operation first. As it'll be a much simpler change to test once it's scoped to that.

netkan/netkan/cli/utilities.py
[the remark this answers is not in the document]
techman83: I've already done this in a separate PR. As there are some logistics in deployment of the changes, I'd rather do them separately. I've already written the code, and will rebase after this is merged.

netkan/netkan/common.py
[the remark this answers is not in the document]
techman83: Because we don't actually install the types for the deployment, they're entirely unnecessary. It allows us to deploy the code, whilst retaining the ability to type check during test/development. It is a nuance of the way type checking has been introduced to Python, and not an uncommon pattern.

netkan/netkan/common.py
techman83: Agreed.

netkan/netkan/common.py
techman83: That's valid, I put them there being a common thing, but it's really more related to queue management.

techman83:
> Yeah, that was very much intentional. Retrieving messages from a queue and deleting them is, frankly, too simple of a task to be worth centralizing. (To me it's a bit like making a wrapper around
logging.error to reduce the number of "implementations" of logging an error message; doing the thing itself instead is simpler and more maintainable.) Especially when it means that key logic that I need to read to understand what it's doing and investigate/fix bugs is now spread across multiple different files and classes rather than in a simple "X then Y then Z" procedural code block.

This decouples the queue handling, so that when debugging an issue or writing test cases, the queue can be largely ignored. They all pretty much were doing the same thing, but three different ways. I still think there is definitely room for simplifying further, but it works, has tests, and is relatively clean.

techman83:
@HebaruSan - any other blockers here? #288 + #289 are considerably more tightly scoped, so they should be much less to get across after this one.

netkan/netkan/indexer.py
HebaruSan: Inheritance would be handled thusly in the derived class:

```python
    def __init__(self) -> None:
        super().__init__(params)
```

(In most languages the compiler would do this for you, but Python is special. :snake:)

HebaruSan:
I reckon the unnecessary lazy initialization is a enough of a code smell to be a blocker. It's extra code we don't need, and it sends a misleading signal to the person reading the code.

netkan/netkan/indexer.py
techman83: I realise I didn't give a complete answer here. Mostly because I do a lot of this in the code I write at work. 

Type safety is the predominant reason, it means we can get rid of the Optional. The static definition at the top of the class only defines the type when you don't set it, hence the need to "lazy load".

It's my preference to be strict about what a property has, rather than needing to worry about it silently none (as far as the type checker is concerned).

techman83:
It's not really a code smell, rather a pattern that's an artifact of a "typed" runtime language. Personally I'd rather see see coffee battle tested than spend too much time worrying about code smells. It has tests, we can refactor if we find a better way to achieve the same goal.

HebaruSan:
Can we please just instantiate those properties in the `__init__` like normal?

netkan/netkan/indexer.py
HebaruSan: What I'm suggesting is perfectly type safe and doesn't require Optional. Sounds like a good reason not to use those static definitions.

techman83: @HebaruSan oh, I confused this with something else I was looking at. Doing it this way avoids the need to override the `__init__` in the consuming class (due to the inherent type challanges, it does make it a consistent pattern with the challenge I mentioned). Which is generally preferable in Python land. The static analysis tools I use at work (paid) haven't ever marked this pattern as a code smell and pylint hasn't complained. There is nothing here that is inherently wrong beyond a difference in style preferences. I don't feel like it's worth refactoring further in this PR. If we're really motivated we can  circle back around with a better pattern if desired, as there are test cases.

techman83:
At this point I'd rather not make any further changes to the PR. Ultimately it's a pretty common python pattern, and the decision to do it this way were unrelated to any thought towards performance.

netkan/netkan/indexer.py
HebaruSan: It's desired! Strongly desired!

HebaruSan:
I guess I might as well stop reviewing code, then, if the possibility of changes has been ruled out.

netkan/netkan/indexer.py
techman83: The alternative pattern to this would be to write custom descriptors or use a dataclass. Which for this use case seems *way* overkill to make this a little more "DRY", along with burying the logic further. I personally find a class more readable, and the added ability to inline docstrings when the properties aren't all buried up in the init. It is notable that the complaints from our static analyses tools and heavy use of typing have driven me towards these patterns.

If it was purely an optimisation thing, I'd 100% be onboard with what you're saying.

techman83:
Ultimately it's a stylistic choice and I'm being stubborn. On reflection if that makes it easier and gets it across the line, I'm good with that.

HebaruSan (approved):
Let's goooo! 🏃
(I'll leave this for you to merge, so you can monitor it during deployment and rebase other PRs as needed)

HebaruSan:
> Especially when it means that key logic that I need to read to understand what it's doing and investigate/fix bugs is now spread across multiple different files and classes rather than in a simple "X then Y then Z" procedural code block.

A prophet, me? If you say so... 😁

HebaruSan:
> It's not really a code smell

Here's an explanation of what "code smell" means:

- <https://www.youtube.com/watch?v=4gXbAday4Og>

It's doesn't refer (only) to a kind of warning from a linter. It's when "part of a program is not ideally structured, not necessarily an error or a bug". Examples given are:

- long method
- long parameter list
- large class
- lazy class
- duplicated code
- message chains

"Extra code" (5 lines to do 1 line of work) and "unmotivated lazy initialization" are in the same category of concerns.

PR #284, opened 2023-03-11
Add KSP2 Inflator

techman83:
This adds a secondary inflator and queue for ksp2 inflations, along with adding the relevant GAME env var to them.

- No changes to existing inflator infra, beyond exposing the GAME environment variable
- Secondary inflator will sit dormant until the rest of the pieces are added

This should allow us to get KSP-CKAN/CKAN#3797 merged

### DIFF
--- a/prod-stack.py
+++ b/prod-stack.py
@@ -48,6 +48,10 @@ inbound = t.add_resource(Queue("NetKANInbound",
                                QueueName="Inbound.fifo",
                                ReceiveMessageWaitTimeSeconds=20,
                                FifoQueue=True))
+inbound2 = t.add_resource(Queue("NetKANKSP2Inbound",
+                                QueueName="InboundKsp2.fifo",
+                                ReceiveMessageWaitTimeSeconds=20,
+                                FifoQueue=True))
 outbound = t.add_resource(Queue("NetKANOutbound",
                                 QueueName="Outbound.fifo",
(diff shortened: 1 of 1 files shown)


prod-stack.py
```diff
@@ -754,6 +760,27 @@
...
+            ('GAME', 'KSP')
+        ],
+        'volumes': [
+            ('ckan_cache', '/home/netkan/ckan_cache')
+        ]
+    },
+    {
+        'name': 'InflatorKsp2',
+        'image': 'kspckan/inflator',
+        'memory': '256',
```
HebaruSan, line 772: Looks like the original Inflator had `156` before #224, we could probably start at least that low here to save memory given that both repos are currently <200KiB and the downloads are at least that tiny with no part or planet packs yet:

```
$ du -sh KSP2*
153K    KSP2-CKAN-meta
130K    KSP2-NetKAN
```

HebaruSan (approved):
Pending merge of KSP-CKAN/CKAN#3797 and consideration of comment about RAM, this is what we need as far as I understand it.

prod-stack.py
techman83: True, though it's the max memory they're allowed to use rather then allocated. There is just shy of 900MB headroom for it, so we should be ok.

HebaruSan: Works for me!

PR #285, opened 2023-03-11
fix: Redeploy Perms

techman83:
The permissions for re-deploying services were missing the renamed inflator and the new inflator, this addresses that and allows for deployment to proceed.


HebaruSan (approved):
Ah yes, I forgot to check the lists of service names as well. Good catch!

PR #288, opened 2023-03-27
Refactor Git Repos

techman83:
Many assumptions around branches and repos were made early on, since then we moved a lot of the common repo actions into relevant classes. This furthers that work and replaces all direct branch access with methods that automatically figure out the primary branch of the repo. Tests updated to reference 'main'

It is intended that #283 is reviewed and merged first, then once happy we can move onto this in preparation for adding KSP2 config. We will be able to follow soon after by renaming master -> main for the existing KSP repos.

### DIFF
--- a/netkan/netkan/auto_freezer.py
+++ b/netkan/netkan/auto_freezer.py
@@ -16,7 +16,7 @@ class AutoFreezer:
         self.github_pr = github_pr
 
     def freeze_idle_mods(self, days_limit: int, days_till_ignore: int) -> None:
-        self.nk_repo.git_repo.remotes.origin.pull('master', strategy_option='ours')
+        self.nk_repo.pull_remote_primary(strategy_option='ours')
         idle_mods = self._find_idle_mods(days_limit, days_till_ignore)
         if idle_mods:
             with self.nk_repo.change_branch(self.BRANCH_NAME):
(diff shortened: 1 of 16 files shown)


netkan/netkan/spacedock_adder.py
```diff
@@ -71,11 +73,28 @@ def try_add(self) -> bool:
...
+            # Commit - Use Submitter username/email
+            self.nk_repo.git_repo.index.commit(
+                (
+                    f"Add {self.info.get('name')} from {self.info.get('site_name')}"
+                    f"\n\nThis is an automated commit on behalf of {self.info.get('username')}"
+                ),
+                author=git.Actor(self.info.get('username'),
+                                 self.info.get('email'))
+            )
+            self.nk_repo.push_remote_branch(branch_name)
```
HebaruSan, line 91:
```suggestion
```

Bad merge, this block shouldn't be here (deleted in #291, we use `with self.nk_repo.change_branch` now).

HebaruSan (changes requested):
One comment about resurrected code in SpaceDockAdder aside, this looks solid and ready to merge. 🎉

HebaruSan (approved):
Looks good, let's go with this. 🏍️

techman83:
Resolved the bad merge conflict resolution, and removed all the reformatting. I think I'd like to switch to black for the project, but I'd rather not mix in formatting with refactoring.

PR #289, opened 2023-03-29
Automatically Iterate Games + Add KSP2 Config

techman83:
This PR is intended to come after #283 + #288 - will rebase when ready

## Adds
- `games` method that automatically returns a list of the configured games
- KSP2 Configuration

## Changes
- Utilities configured within the stack will iterate through all available games (Scheduler/Freezer/Counter)

## Fixes
- Remove `processed` from the QueueHandler, avoiding the risk of the list growing perpetually until SQS throws an exception

## Extra
- Left the mirrorer out to focus on it as a separate piece in preparation for when we have > 50 mod versions

### DIFF
--- a/netkan/netkan/cli/common.py
+++ b/netkan/netkan/cli/common.py
@@ -182,6 +182,7 @@ class SharedArgs:
     user: str
     _debug: bool
     _ssh_key: str
+    _game_ids: List[str]
 
     def __init__(self) -> None:
         self._environment_data = None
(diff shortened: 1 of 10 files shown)


techman83:
@HebaruSan - this will still require rebasing of the git-repos work, but I have reabsed off master to make it a bit easier to review

netkan/netkan/cli/common.py
```diff
@@ -227,6 +228,18 @@ def game(self, game: str) -> Game:
             setattr(self, f'_game_{game_id}', Game(game_id, self))
         return getattr(self, f'_game_{game_id}')
 
+    @property
+    def games(self) -> List[str]:
```
HebaruSan, line 232: Can we call this `game_ids` if it's going to return strings, or have it return `Game`s if it's going to be called `games`?

techman83: Yep, thanks for keeping on top of the consistency!

HebaruSan (approved):
Yeah, looks fine. Let's see if it works! 🚀

PR #290, opened 2023-04-06
Ensure processed queue is emptied

techman83:
As the handlers are kept in memory, they keep this list state between processing. This leads to the occaisional failure when we inevitably end up with more than 10 in the list. We should consider tests and a cleaner solution, but this will keep the bots uptime and logs quieter for now.


HebaruSan (approved):
Obviously I hate this, but let's just hack it till it works and then refactor later

techman83:
@HebaruSan - agreed. Best I could do in 10 minutes :slightly_smiling_face: - added a test as well.

PR #292, opened 2023-04-08
fix: Avoid empty call to SQL delete

techman83:
We always had messages to delete previously, now it's possible/likely to have a game that doesn't have any messages to delete.


techman83:
Self reviewing. We now have mutli games, so it's possible to encounter this scenario!

PR #293, opened 2023-04-08
Log SpaceDockAdder._pr_body input on failure

HebaruSan:
## Problem

This happened:

```
Uncaught exception:
Traceback (most recent call last):
  File ".local/bin/netkan", line 8, in <module>
    sys.exit(netkan())
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/cli/services.py", line 75, in spacedock_adder
    SpaceDockAdderQueueHandler(common).run()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/queue_handler.py", line 98, in run
    processed = handler.process_messages()
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/spacedock_adder.py", line 164, in process_messages
    return [c.delete_attrs for c in self._process_queue(self.queued)]
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/spacedock_adder.py", line 159, in _process_queue
    if netkan.try_add():
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/spacedock_adder.py", line 80, in try_add
    body=SpaceDockAdder._pr_body(self.info),
  File "/home/netkan/.local/lib/python3.7/site-packages/netkan/spacedock_adder.py", line 90, in _pr_body
    for a in [info, *info.get('shared_authors', [])])}))
[...]

PR #294, opened 2023-04-08
Handle bad author list input from SpaceDock

HebaruSan:
## Problem

See #293, the SpaceDockAdder is repeatedly failing for a mod with co-authors.

## Cause

After #293, we can see that SpaceDock is passing a string where it's supposed to pass a list of dicts:

```json
"shared_authors": "username"
```

(`username` is the first key in the contained dicts)

## Changes

Now the SpaceDockAdder performs run-time type checks on the shared authors property and only uses it to generate the PR body if it's correct. Otherwise an error is logged.

This will allow the bot to finally finish processing the request that's been stuck in the queue. The underlying SpaceDock bug will be fixed later.

### DIFF
--- a/netkan/netkan/spacedock_adder.py
+++ b/netkan/netkan/spacedock_adder.py
@@ -84,11 +84,19 @@ class SpaceDockAdder:
     @staticmethod
     def _pr_body(info: Dict[str, Any]) -> str:
         try:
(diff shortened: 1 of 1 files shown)


HebaruSan:
Worked! KSP-CKAN/KSP2-NetKAN#17

image

PR #295, opened 2023-04-12
Multigame ModStatus

techman83:
This is an initial piece of work that will at least set the 'game_id' in our status table in advance of moving the query to a Hash + Range key.

No changes to the Table will happen in this PR. Upon merge, status updates will begin populating the field.


HebaruSan (approved):
Makes sense so far...

PR #298, opened 2023-04-15
Multi Game Status

techman83:
This changes the dynamodb table to us a Range key of 'game_id', so that we can query a record specifically by ModIdentifier + game_id.

It includes the ability to export/import all mods, defaulting the game_id to ksp on import. This is required as setting a range key requires recreation of the table.

Export to S3 splits the games on export. Persisting how the current status works, giving us some time to address it separately.

I've created/done an initial load of the data to the new table, so if all looks good, we can merge and handle the clean up of the previous table after.

### DIFF
--- a/netkan/netkan/status.py
+++ b/netkan/netkan/status.py
@@ -3,10 +3,12 @@ import logging
 import os
 import time
 from datetime import datetime, timezone
+from itertools import chain
 from pathlib import Path
 from typing import Optional, Dict, Any
 from dateutil.parser import parse
 from pynamodb.models import Model
+from pynamodb.expressions.condition import Condition
 from pynamodb.attributes import (
     UnicodeAttribute, UTCDateTimeAttribute, BooleanAttribute, MapAttribute
 )
@@ -22,7 +24,7 @@ from .repos import CkanMetaRepo
 
 
 def table_name() -> str:
-    return os.getenv('STATUS_DB', 'NetKANStatus')
+    return os.getenv('STATUS_DB', 'MultiKANStatus')
 
 
 def region() -> str:
@@ -35,7 +37,7 @@ class ModStatus(Model):
         region = region()
 
     ModIdentifier = UnicodeAttribute(hash_key=True)
-    game_id = UnicodeAttribute(null=True)
+    game_id = UnicodeAttribute(range_key=True)
     last_error = UnicodeAttribute(null=True)
     last_warnings = UnicodeAttribute(null=True)
     last_checked = UTCDateTimeAttribute(null=True)
@@ -65,9 +67,9 @@ class ModStatus(Model):
     # So we'd probably need to be tracking 10,000+ mods before it becomes
     # a problem.
     @classmethod
-    def export_all_mods(cls, compat: bool = True) -> Dict[str, Any]:
-        data = {}
-        for mod in cls.scan(rate_limit=5):
+    def export_game_mods(cls, condition: Condition, compat: bool = True) -> Dict[str, Any]:
+        data: Dict[str, Any] = {}
+        for mod in cls.scan(rate_limit=5, filter_condition=condition):
             data[mod.ModIdentifier] = mod.mod_attrs()
 
             # Persist compability with existing status ui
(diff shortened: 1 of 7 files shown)


netkan/netkan/status.py
```diff
@@ -79,50 +81,64 @@ def export_all_mods(cls, compat: bool = True) -> Dict[str, Any]:
         return data
 
     @classmethod
-    def export_to_s3(cls, bucket: str, key: str, compat: bool = True) -> None:
+    def export_all_mods(cls, compat: bool = True) -> Dict[str, Any]:
+        data: Dict[str, Any] = {}
+        for game_id in ['ksp', 'ksp2']:
```
HebaruSan, line 86: Is there any way we could get the list of game ids from the command line or the db, instead of hard-coding it?

netkan/netkan/status.py
```diff
@@ -79,50 +81,64 @@ def export_all_mods(cls, compat: bool = True) -> Dict[str, Any]:
...
+            for game in chain(existing.values()):
+                for key, item in game.items():
+                    # Every batch write consumes a credit, we want to leave spare
+                    # credits available for other operations and also not error out
+                    # during the operation (pynamodb doesn't seem to have a limit
+                    # option on batch queries).
+                    if len(batch.pending_operations) == 5:
+                        batch.commit()
+                        time.sleep(1)
+                    cls.normalise_item(key, item)
```
HebaruSan, line 140: - In the old code, we normalize first and commit batch second
- In the new code, we commit batch first and normalize second

Just wanted to double check that this change is OK and intentional, since it looks like it could be something that happens accidentally in the course of refactoring.

prod-stack.py
```diff
@@ -79,24 +79,33 @@
         ),
     ])
 
-INFLATION_QUEUES = Sub('ksp=${ksp} ksp2=${ksp2}', ksp=GetAtt(inbound, 'QueueName'), ksp2=GetAtt(inbound2, 'QueueName'))
+INFLATION_QUEUES = Sub('ksp=${ksp} ksp2=${ksp2}', ksp=GetAtt(
+    inbound, 'QueueName'), ksp2=GetAtt(inbound2, 'QueueName'))
```
HebaruSan, line 83:
```suggestion
INFLATION_QUEUES = Sub('ksp=${ksp} ksp2=${ksp2}',
                       ksp=GetAtt(inbound, 'QueueName'),
                       ksp2=GetAtt(inbound2, 'QueueName'))
```

Suggest indenting this in a way that more accurately reflects the semantic level grouping of things.

netkan/netkan/status.py
```diff
@@ -79,50 +81,64 @@ def export_all_mods(cls, compat: bool = True) -> Dict[str, Any]:
...
-    def export_to_s3(cls, bucket: str, key: str, compat: bool = True) -> None:
+    def export_all_mods(cls, compat: bool = True) -> Dict[str, Any]:
+        data: Dict[str, Any] = {}
+        for game_id in ['ksp', 'ksp2']:
+            data.update({game_id: cls.export_game_mods(
+                condition=ModStatus.game_id == game_id, compat=compat)})
+        data.update({'no_game_id': cls.export_game_mods(
+            condition=ModStatus.game_id.does_not_exist(), compat=compat)})
+        return data
+
```
HebaruSan, line 92: On second thought, this isn't called anywhere. Maybe just delete it if we don't need it?

HebaruSan, line 92:
```suggestion
```

On second thought, this isn't called anywhere. Maybe just delete it if we don't need it?

(Sorry for the notification spam, having a bit of a struggle with the GH web UI here)

HebaruSan (approved):
No showstoppers here, ~~but `export_all_mods` is struggling to provide an operation we don't need. If we just delete it, we can invest more effort into making it better someday if we find out we do need it.~~

netkan/netkan/status.py
techman83: Sort of, but not really. We'd have to scan through all the mods, then split them up. I kinda wanted to re-use the same export function as S3 as for the table dump/restore functions, so this was the middle ground.

netkan/netkan/status.py
techman83: It's used by the dump_status utility, which will be involved in how the tables are transitioned from having no range key. Which is a destructive update (you can create secondary indexes, but I think for simplicity and the tiny amount of data, migrating makes sense).

All good, happy with the questions.

netkan/netkan/status.py
techman83: It was intentional, probably not super necessary. But realistically this data wasn't getting used until the batch.save() when the status is getting instantiated. It felt more sensible to do it when necessary, rather then before the commit.

prod-stack.py
techman83: That was autopep8 and I didn't notice it. I think switching to black will be a good idea. It's jarring at first, but it's consistent and I hate getting tied up thinking about how best to format things. Python has pretty thorough standards around it, even if I don't like all of them.

techman83:
Awesome, I'll do a final dump/restore of the tables and get this merged. `export_all_mods` is very much involved in that process, it's also useful for repopulating the dev table.

netkan/netkan/status.py
HebaruSan: Databases typically have indexing capabilities that can handle this kind of operation very quickly. My guess is it's somewhere in pynamodb as well.

netkan/netkan/status.py
HebaruSan: Ahh, I see. From the way the diff looked I took this to be a new function, and ctrl-F didn't find it anywhere else.

netkan/netkan/status.py
techman83: DynamoDB isn't really a traditional database, so that's not quite how the indexing works. So you have to scan all the entries and find the uniqueness client side.

PR #299, opened 2023-04-17
Indexer Status

techman83:
The range key was missed in the gets, this fixes the indexer to unblock that function


techman83:
I'll self review to unblock the indexing. The autofreezer I have a fix for, but the test case needs some work.

PR #286, opened 2023-03-15
Raise error on failed push

HebaruSan:
I have been trying to figure out why #281 didn't work, and I'm stumped.

The push must be failing, but it's silently ignored. Now it'll throw an exception on failure, so we can get more clues about why we never see the AutoFreezer's additional commits when the remote branch exists.

<https://gitpython.readthedocs.io/en/stable/tutorial.html#handling-remotes>

PR #300, opened 2023-04-17
AutoFreezer ModStatus

techman83:
As with the indexer, the range_key was overlooked. This ensures the game_id is passed to ModStatus.get, which is compliant with the new schema.

### DIFF
--- a/netkan/netkan/auto_freezer.py
+++ b/netkan/netkan/auto_freezer.py
@@ -11,9 +11,10 @@ class AutoFreezer:
 
     BRANCH_NAME = 'freeze/auto'
 
-    def __init__(self, nk_repo: NetkanRepo, github_pr: GitHubPR) -> None:
+    def __init__(self, nk_repo: NetkanRepo, github_pr: GitHubPR, game_id: str) -> None:
         self.nk_repo = nk_repo
         self.github_pr = github_pr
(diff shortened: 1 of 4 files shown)


HebaruSan (approved):
Looks fine, thanks for updating it.

HebaruSan:
BTW, while we're looking at this class, any ideas why the push is still failing when the remote branch already exists? As far as I can tell it's using `XkanRepo.change_branch` the same as the Indexer and Adder, but with different results (see https://github.com/KSP-CKAN/NetKAN-Infra/pull/286#issuecomment-1471267161). I'd love be able to defer reviewing AutoFreezer PRs without losing commits.

techman83:
> BTW, while we're looking at this class, any ideas why the push is still failing when the remote branch already exists? As far as I can tell it's using XkanRepo.change_branch the same as the Indexer and Adder, but with different results (see #286 (comment)). I'd love be able to defer reviewing AutoFreezer PRs without losing commits.

I had hoped that aligning that with the rest of how the repos does it, that it wouldn't occur anymore. If it's still occurring I can look closer. Maybe open an issue and we can track it a bit better.

HebaruSan:
> Maybe open an issue and we can track it a bit better.

Good idea, done in #302.

PR #301, opened 2023-04-17
Ensure Status range_key is consistent

techman83:
All range keys are lower, however we could be asking for it in upper, which won't exist and upon save PynamoDB will dutifly wipe out the fields that aren't set. Instead of doing the update during status as expected.


techman83:
Self reviewing to fix clobbering of existing data.

PR #303, opened 2023-04-21
Auto Freezer

### LINKED ISSUE #302: AutoFreezer fails to re-use branch and PR
## Problem

The AutoFreezer runs once a week, pushes a branch, and creates a pull request.

If we don't merge the PR and delete the branch within a week, the next attempt fails. If this happens 3 weeks in a row, then some mods can pass through the window where they're idle enough to be considered (1000 days) but not idle enough to be skipped as permanently active (1021 days) without being reviewed. We've had a few such spans now, so we'll have to do some catch-up eventually.

### History

I started trying to grapple with this in #281; I thought the problem was we weren't pulling the branch before switching to it, but switching to `XkanRepo.change_branch` didn't help. The push fails, even though this exact scenario works fine for the Indexer and the Adder.

Then #286 started logging stack traces for failed pushes, which didn't tell us anything.

techman83:
This is a tweak to the auto-freezer, as we're hitting some different limits as KSP ages, and more effort heading to KSP2

## Fixes
- Force a deep clone, fixing branch switching
- Commit and pause during batched freezing to avoid WCU limits of DynamoDB
- Freeze process not filtering for game_id

## Changes
- Increase memory for the auto freezer

Fixes #302.

### DIFF
--- a/netkan/netkan/auto_freezer.py
+++ b/netkan/netkan/auto_freezer.py
@@ -1,3 +1,4 @@
+import time
 import logging
 from datetime import datetime, timezone, timedelta
 from typing import Iterable, Optional, List, Tuple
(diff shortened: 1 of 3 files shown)


HebaruSan (approved):
Nice work! I would still like to know why the bot's pull request comment never appears, though.

techman83:
Added some extra logging for the comment.

PR #304, opened 2023-04-21
PR Comment

techman83:
We were attempting to use the string representation of the owner object. This may have returned something that worked in the past, but now doesn't.


techman83:
Fairly minor change, self reviewing so we can call it done :slightly_smiling_face:

PR #306, opened 2023-08-09
Handle multiple download URLs per module

HebaruSan:
## Motivation

As of KSP-CKAN/CKAN#3877 and the next release of the CKAN client, it will be possible for the `download` property to contain a list of multiple URLs. A few places in the Infra assume it will be a single value.

## Changes

- The existing `Ckan.download` property is updated to preserve backwards compatibility by always returning the first URL if there are multiple. This will ensure that all old code using this property will continue to work as previously. For example, the `Mirrorer` doesn't really need to check multiple URLs to do its job, so it will continue using this property.
- A new `Ckan.downloads` property is added that always returns a list, which includes the archive.org fallback URLs for modules that have them. The `DownloadCounter` is updated to use this property to count and sum all of the downloads for modules that have multiple URLs.

### DIFF
--- a/netkan/netkan/download_counter.py
+++ b/netkan/netkan/download_counter.py
@@ -113,10 +113,7 @@ class GraphQLQuery:
             if user_repo in self.cache:
                 count = self.cache[user_repo]
                 logging.info('Count for %s is %s', ident, count)
-                if ident in counts:
-                    counts[ident] += count
-                else:
-                    counts[ident] = count
+                counts[ident] = counts.get(ident, 0) + count
         return counts
 
     def graphql_to_github(self, query: str) -> Dict[str, Any]:
@@ -171,10 +168,7 @@ class SpaceDockBatchedQuery:
             count = sd_counts.get(sd_id)
             if count:
                 logging.info('Count for %s is %s', identifier, count)
-                if identifier in counts:
-                    counts[identifier] += count
-                else:
-                    counts[identifier] = count
+                counts[identifier] = counts.get(identifier, 0) + count
         return counts
 
 
@@ -207,10 +201,7 @@ class InternetArchiveBatchedQuery:
         result = requests.get(self.IARCHIVE_API + ','.join(self.ids.values()),
                               timeout=60).json()
         for ckan_ident, ia_ident in self.ids.items():
-            if ckan_ident in counts:
-                counts[ckan_ident] += result[ia_ident]['all_time']
-            else:
-                counts[ckan_ident] = result[ia_ident]['all_time']
+            counts[ckan_ident] = counts.get(ckan_ident, 0) + result[ia_ident]['all_time']
         return counts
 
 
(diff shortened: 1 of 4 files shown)


HebaruSan:
I don't know what's wrong with the Coverage test, but it doesn't look like it's related to these changes.

netkan/netkan/download_counter.py
```diff
@@ -113,7 +113,10 @@ def get_result(self, counts: Optional[Dict[str, int]] = None) -> Dict[str, int]:
             if user_repo in self.cache:
                 count = self.cache[user_repo]
                 logging.info('Count for %s is %s', ident, count)
-                counts[ident] = count
+                if ident in counts:
+                    counts[ident] += count
+                else:
+                    counts[ident] = count
```
techman83, line 119: I always feel a bit funny about these kinds of if/else statements in python (and not using dict methods), this would remove the nested if/else and achieve the same outcome.
```suggestion
                counts.update({ident: counts.get(ident, 0) + count})
```

```python
In [1]: ident = 'AwesomeMod'

In [2]: counts = {}

In [3]: count = 3

In [4]: counts.update({ident: counts.get(ident, 0) + count})

In [5]: counts
Out[5]: {'AwesomeMod': 3}

In [6]: counts.update({ident: counts.get(ident, 0) + count})

In [7]: counts
Out[7]: {'AwesomeMod': 6}
```

HebaruSan: I don't love the if/else either, but I like it better than having to know what `update` does and creating a throwaway `dict` object.

What if we refactor the if/else into an `AddOrSum` method?

HebaruSan: Found a compromise that does it in one statement without using `update` and an extra `dict`.

netkan/netkan/download_counter.py
```diff
@@ -168,7 +171,10 @@ def get_result(self, counts: Optional[Dict[str, int]] = None) -> Dict[str, int]:
             count = sd_counts.get(sd_id)
             if count:
                 logging.info('Count for %s is %s', identifier, count)
-                counts[identifier] = count
+                if identifier in counts:
+                    counts[identifier] += count
+                else:
+                    counts[identifier] = count
```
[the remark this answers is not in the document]
HebaruSan: Found a compromise that does it in one statement without using `update` and an extra `dict`.

netkan/netkan/download_counter.py
```diff
@@ -201,7 +207,10 @@ def get_result(self, counts: Optional[Dict[str, int]] = None) -> Dict[str, int]:
         result = requests.get(self.IARCHIVE_API + ','.join(self.ids.values()),
                               timeout=60).json()
         for ckan_ident, ia_ident in self.ids.items():
-            counts[ckan_ident] = result[ia_ident]['all_time']
+            if ckan_ident in counts:
+                counts[ckan_ident] += result[ia_ident]['all_time']
+            else:
+                counts[ckan_ident] = result[ia_ident]['all_time']
```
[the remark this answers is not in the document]
HebaruSan: Found a compromise that does it in one statement without using `update` and an extra `dict`.

techman83 (approved):
Awesome, thanks @HebaruSan !

techman83:
We can fix the coverage separately, it's failing for issues unrelated to this PR

PR #307, opened 2023-08-31
Don't allow spaces in filenames for archive.org

HebaruSan:
## Problem

image

## Cause

Spaces in versions wreaking havoc again.

## Changes

Now spaces in versions are replaced with underscores when uploading to archive.org. Will need a corresponding change in the client to actually use such fallbacks.

Will self-review.


HebaruSan:
Coverage still not working, rest looks fine.

HebaruSan:
https://archive.org/details/MinotaurPack-0.1.0_Beta

PR #315, opened 2023-09-05
Use deep clone for SpaceDockAdder

HebaruSan:
## Problem

We've been getting spammed by these every 5 minutes:

image

## Cause

The CKAN checkbox came unchecked for this mod on SpaceDock, so I re-checked it, which caused the webhooks to put an entry into the `SpaceDockAdder`'s input queue for it.

The `SpaceDockAdder` then fails to process it, leaves it in the queue, and tries again in 5 minutes. This happens because KSP-CKAN/NetKAN#9778 is still open and its branch still exists, so the adder fails to push the branch, because it's using a shallow clone.

This is the same problem that the `AutoFreezer` was having in #302.

## Changes

Just like #303 did for the `AutoFreezer`, the `SpaceDockAdder` now uses a deep clone, which will allow it to re-use existing branches.

### DIFF
--- a/prod-stack.py
+++ b/prod-stack.py
@@ -842,7 +842,7 @@ services = [
     },
     {
         'name': 'Adder',
-        'command': 'spacedock-adder',
+        'command': ['spacedock-adder', '--deep-clone'],
         'secrets': ['GH_Token', 'SSH_KEY'],
         'env': [
             ('SQS_QUEUE', GetAtt(addqueue, 'QueueName')),


HebaruSan:
Didn't work. ☹️

techman83:
I was going to do this, but was feeling a bit under the weather. Thanks @HebaruSan - I'll nudge this out shortly!

HebaruSan:
Oh no, get well soon! 🤒

PR #316, opened 2023-09-08
Fewer mods per GraphQL pass in DownloadCounter

HebaruSan:
## Problem

These have been getting more and more common lately:

image

A mod author asked about the download counts changing unexpectedly in another Discord.

## Cause

The cause is impossible to identify with certainty because the error message is so absurdly, comically vague—"Something"? "this _may_ be"??—but [some reports online indicate that it may indeed be due to a timeout](https://github.com/orgs/community/discussions/24631) caused by too complex of a query, i.e. the amount of work we can do is arbitrarily limited per arbitrary bucket.

## Changes

Now the download counter will send GraphQL requests for 20 mods at a time instead of 40. Since we don't and can't know what the problem is, it's impossible to guess whether this will change anything, but if the cause is in fact a timeout 🤞, then this should give our queries a bit more breathing room to finish before they are speculatively terminated by a possible time limit of unknown nature or even existence. (Never mind that the overall amount of work done is exactly the same, just split up into a greater number of smaller groups. 🙄)

Also the broken Coverage check is now disabled in this repo with an `if: false` condition because I got tired of seeing the error messages. We can turn it back on once it's working again.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -8,6 +8,7 @@ permissions:
 
 jobs:
   coverage:
+    if: false
     runs-on: ubuntu-latest
     defaults:
       run:
--- a/netkan/netkan/download_counter.py
+++ b/netkan/netkan/download_counter.py
@@ -20,7 +20,7 @@ class GraphQLQuery:
     GITHUB_API = 'https://api.github.com/graphql'
 
     # Get this many modules per request
-    MODULES_PER_GRAPHQL = 40
+    MODULES_PER_GRAPHQL = 20
 
     # The request we send to GitHub, with a parameter for the module specific section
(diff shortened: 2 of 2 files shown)


techman83:
Thanks @HebaruSan - annoying that immediately after implementing that, it broke. I have work to do on some other projects, so I'll have a play around with it.

HebaruSan:
Yeah, it's surprising how often things like that just break, given all the work that goes into making them work in the first place.

Maybe the easiest thing would be to start over and find a brand new coverage checker, rather than trying to fix this one?

techman83:
Well the coverage tool is fine, most of the projects use it to build reports from. It's just the reporting action that is a problem.

PR #317, opened 2023-10-28
Add Aspect to ModAnalyzer to detect patch manager patches

cheese3660:
## Motivation
A mod has been released for KSP2, `PatchManager`, that uses special files (.patch files) to either patch data in the addressables for the game, and/or add new data to the addressables assets. any KSP2 mod that use files with this extension necessarily depend on this mod, examples being `CommunityResources`, or `KerbalLifeSupportSystem`. So it would make sense for them to be autodetected to add `PatchManager` as a dependency to the mods via the `ModAnalyzer` class.

## Changes
Adds an aspect to the `ModAnalyzer` class that detects *.patch files, and adds the `config` tag (closest tag to match for what patch files are), and adds a dependency on `PatchManager`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,14 +2,6 @@
   "python.pythonPath": "/usr/bin/python3",
   "python.formatting.provider": "autopep8",
   "editor.formatOnSave": true,
-  "python.linting.pylintEnabled": true,
-  "python.linting.enabled": true,
-  "python.linting.pylintArgs": [
-    "--rcfile",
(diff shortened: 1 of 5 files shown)


HebaruSan (approved):
Obviously this code change is correct, but I'm going to dig into why the tests are emitting errors now... (This happens regularly; dependencies or Python tools get updated and new errors appear where previously there were none, and we try to keep up with these as new code changes are made so they don't accumulate.)

PR #318, opened 2023-11-13
Fix timezone problem in ticket closer

HebaruSan:
## Problem

The ticket closer has thrown this the last 2 days:

image

## Cause

Python throws a runtime exception if time zones aren't handled to its liking. 🐍 

`datetime.now()` doesn't have timezone info associated with it, so by process of elimination `issue.updated_at` must have it. Maybe we updated the GitHub API library recently. 🤷 

## Changes

Now we pass `timezone.utc` to `datetime.now()` so both parts of the comparison will have timezone data.

I'll self review this so we can see if it works in tomorrow's run.

### DIFF
--- a/netkan/netkan/ticket_closer.py
+++ b/netkan/netkan/ticket_closer.py
@@ -1,5 +1,5 @@
 import logging
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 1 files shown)


techman83:
Thanks for fixing this @HebaruSan !

PR #320, opened 2023-12-03
ticket_closer should set the state_reason to not_planned

JonnyOThan:
api reference here: https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#update-an-issue


HebaruSan (approved):
Yeah, that makes sense. Thanks!

(I think this was originally written before GitHub introduced the reasons for closing.)

PR #322, opened 2023-12-16
Update to Python 3.11

HebaruSan:
Python 3.7 is getting a little long in the tooth, and we've started to receive warnings about it no longer being supported during the builds. Let's see how the workflows handle 3.11.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7
+          python-version: 3.11
       - name: Install test dependencies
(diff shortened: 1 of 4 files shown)


techman83 (approved):
I reckon let's give this a shot, I've been using similar packages across 3.8 - 3.11 without too many issues. Should also come with some performance / memory benefits too.

Thanks @HebaruSan !

HebaruSan:
Note that this still leaves the meta tester with whatever Python3 is in `mono:latest`, so we'll have to avoid taking on a dependency on later language versions until we figure that out.

https://github.com/KSP-CKAN/CKAN/blob/d43a2ac68f119cad2909353339dd7a0dafb9ff1c/Dockerfile.metadata#L6

PR #323, opened 2023-12-19
Prune Docker image

HebaruSan:
The `kspckan/netkan` image inherits a lot of cruft from `python:3.11`: autoconf, automake, make, dpkg-dev, gcc, g++, hicolor-icon-theme, imagemagick, subversion, tcl/tk, lots of libs, x11-common, etc., as well as the package download cache in `/home/netkan/.cache/pip`. This bloats it quite a bit:

```
REPOSITORY                         TAG       IMAGE ID       CREATED              SIZE
kspckan/netkan                     latest    4883cd42823b   3 days ago           1.11GB
```

Now it starts from `ubuntu:latest`, installs only the Python components it needs for each step, and excludes the pip cache from the copy-forward of `/home/netkan`, resulting in a much slimmed down image (~31% of previous):

```
REPOSITORY                         TAG       IMAGE ID       CREATED              SIZE
infra_testing                      latest    13fa9e747ffa   About a minute ago   355MB
```

This may slightly speed up redeploying the bot.

Note that this will downgrade Python from 3.11 to 3.10, which will not matter in the short term, but over the long term it will automatically get future Python versions as they are added to `ubuntu:latest` and the image is rebuilt.

### DIFF
--- a/netkan/Dockerfile
+++ b/netkan/Dockerfile
@@ -1,4 +1,18 @@
-FROM python:3.11 as base
+FROM ubuntu:latest as ubuntu_with_python
+
+# Don't prompt for time zone
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Install Git and Python
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        git libffi-dev openssh-client \
+        python3 python-is-python3 \
+    && apt-get clean
+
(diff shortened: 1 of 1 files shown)


techman83 (approved):
Nothing stands out, looks great! Thanks @HebaruSan

PR #325, opened 2024-02-03
Update spacedock_adder.py to also generate the Github download in it's NetKANs

cheese3660:
# Motivation
Since version 1.34 of CKAN, CKAN has supported multiple download links in its NetKANs. And since then most mods for KSP2 have been using this feature. But it is slightly tedious when wrangling a NetKAN generated from the spacedock adder to have to go and find the repo, and copy the correct structure to add the github download link.

# Solution.

When the spacedock_adder recieves a request to add a mod to spacedock it now checks if the `source_link` is a valid github repository that has a release with assets, and generates the github portion of a netkan using this repo, and also tries to deduce what the correct `x_netkan_version_edit` is for said github portion. When it does this it will place this portion of the netkan in front of the spacedock portion of the netkan, in line with how most of the KSP2 netkans in this style have already been structured.

### DIFF
--- a/netkan/netkan/spacedock_adder.py
+++ b/netkan/netkan/spacedock_adder.py
@@ -169,7 +169,7 @@ class SpaceDockAdder:
         props: Dict[str, Any] = {}
         try:
             latest_release = gh_repo.get_latest_release()
-        except: # pylint: disable=broad-except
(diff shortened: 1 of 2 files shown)


HebaruSan (approved):
Alright, let's see how this goes. Thanks for putting this together!

PR #328, opened 2024-02-28
Fully sanitize archive.org bucket names

HebaruSan:
## Problem

The bot has been spamming this every 5 minutes for 2 hours:

image

## Cause

The error message explains it well; only certain characters are allowed in archive.org bucket names, and we currently construct those names from user inputs and don't guarantee that the rules will be followed.

We fixed spaces specifically in #307, since they seemed particularly likely to come up, and it was easier than trying to go all the way and guarantee all bucket names fit the allowed pattern. Surely that's sufficient, right? Nobody needs to put even more obscure characters than that into a **version number**, which is typically of the form `1.2.3.4`. Right?

<https://spacedock.info/mod/2604/Tantares%20Intercolor%20(Recolors)>

image

<https://github.com/KSP-CKAN/CKAN-meta/tree/master/IntercolorTantaresRecolors>

image

## Changes

- Now we use a regex to remove invalid characters from archive.org bucket names
- A few trivial changes are made to quell pylint warnings
  - One license URL was triggering a "line to long" warning, and fortunately it already has a server redirect to a shorter URL

Note that the client will need to be updated in order to actually use these uploads.

I'll self-review this and merge to stop the bot spam.

### DIFF
--- a/netkan/netkan/download_counter.py
+++ b/netkan/netkan/download_counter.py
@@ -2,14 +2,13 @@ import json
 import logging
 import re
 from pathlib import Path
-from importlib.resources import read_text
 from string import Template
 import urllib.parse
 from typing import Dict, Tuple, Any, Optional
 
 import requests
 
-from .utils import repo_file_add_or_changed
+from .utils import repo_file_add_or_changed, legacy_read_text
 from .repos import CkanMetaRepo
 from .metadata import Ckan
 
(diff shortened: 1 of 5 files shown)


HebaruSan:
- <https://archive.org/details/IntercolorTantaresRecolors-5.8.1_Fixes__Additions>

PR #342, opened 2024-09-19
Re-enable coverage

techman83:
Renables coverage, any issues I'll debug and ensure my fork with open PR is up to date.

### DIFF
--- a/.github/workflows/coverage-report.yml
+++ b/.github/workflows/coverage-report.yml
@@ -15,7 +15,7 @@ permissions:
 jobs:
   coverage-report:
     runs-on: ubuntu-latest
-    if: ${{ false && github.event.workflow_run.conclusion == 'success' }}
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     steps:
       - uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


techman83:
Double checked where I worked on this feature and it still works. I have changed to the 'demo' branch, but checking both the feat/demo branches are on the same commits

- https://github.com/CallyCo-io/Cally/pull/29#issuecomment-2360238477
- https://github.com/techman83/coverage/commits/demo/workflow_run
- https://github.com/techman83/coverage/commits/feat/workflow_run

If something doesn't work, I'll fix it.

HebaruSan (approved):
> If something doesn't work, I'll fix it.

Alright, brace for impact...!

HebaruSan:
@techman83, should something have happened with this after merge? If so, then it looks like the first thing to fix will be getting it to run at all:

<https://github.com/KSP-CKAN/NetKAN-Infra/actions/workflows/coverage-report.yml>

image

techman83:
That's expected. The workflow needs to exist in the default branch before it can run. This is to prevent someone skirting security by raising a PR that has a workflow that triggers a workflow within the repo, that would have extra permissions it wouldn't ordinarily get. It's why I had to write the changes in the first place.

PR #343, opened 2024-09-20
Refactor Tests

techman83:
This switches over the unittests to use the autodiscovery method. Along with tidying up the the dev environment. The formatting provider is now configured differently, there are better tools, but changing is a bit more involved and will come in a future PR.


.github/workflows/coverage-report.yml
```diff
@@ -15,15 +15,15 @@ permissions:
 jobs:
   coverage-report:
     runs-on: ubuntu-latest
-    if: ${{ false && github.event.workflow_run.conclusion == 'success' }}
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
HebaruSan, line 18: Has the coverage thing actually been fixed?

HebaruSan (changes requested):
Let's not try to sneak re-enabling the broken coverage check into every unrelated PR 😝

techman83:
I'm working on it 😝 I at least understand why now!

PR #348, opened 2024-10-31
Always report Scheduler abort reasons to Discord

HebaruSan:
## Motivation

The Scheduler is having some outages at the moment, and it's not clear why. We get a few messages in Discord about the GitHub API limit, but then it just goes silent. Meanwhile @techman83 can see stuff like this in the admin screens somewhere:

```
[2024-10-31 00:09:37,857] [INFO    ] Run skipped, below cpu credit target (Current Avg: 91)
[2024-10-31 00:09:38,212] [INFO    ] Run skipped, below cpu credit target (Current Avg: 91)
```

## Changes

- Now all of the Scheduler's abort reasons are upgraded from `logging.info` to `logging.error` so we can tell what's happening without accessing the admin side.
- Now those messages always report which game is being skipped, so we can more easily see which ones are duplicates
- Now those messages also report the limit against which the value is being compared, to give a sense of how bad a problem is and how close it is to self-resolving

### DIFF
--- a/netkan/netkan/scheduler.py
+++ b/netkan/netkan/scheduler.py
@@ -115,10 +115,8 @@ class NetkanScheduler:
         if not dev:
             github_remaining = github_limit_remaining(self.github_token)
             if github_remaining < min_gh:
-                logging.error(
-                    "Run skipped, GitHub API limit nearing exhaustion (Current: %s)",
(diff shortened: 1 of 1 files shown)


HebaruSan:
It's CPU credits again:

image

PR #351, opened 2024-11-25
Absolute entrypoint path to run as Action

HebaruSan:
## Motivation

I'm trying to use the `kspckan/netkan` container in a GitHub Action, see KSP-CKAN/CKAN#4265 for details.

GitHub's docker command uses `--workdir /github/workspace` to override the container's `WORKDIR`. This breaks our `ENTRYPOINT` because it uses a relative path.

- <https://github.com/KSP-CKAN/CKAN/actions/runs/12002494825/job/33454558662>

## Changes

Now the `ENTRYPOINT` contains the full absolute path.


HebaruSan:
A few more wrinkles to iron out, apparently.

image

PR #361, opened 2026-01-10
Add x_netkan_check_parent_downloads

Clayell:
When x_netkan_check_parent_downloads is false, do not check the github parent repos for the download count. Defaults to true. (the current behavior)
Unsure of how to test this to make sure it works.

### DIFF
--- a/netkan/netkan/cli/utilities.py
+++ b/netkan/netkan/cli/utilities.py
@@ -60,6 +60,7 @@ def download_counter(common: SharedArgs) -> None:
         logging.info('Starting Download Count Calculation (%s)...', game_id)
         DownloadCounter(game_id,
                         common.game(game_id).ckanmeta_repo,
+                        common.game(game_id).netkan_repo,
                         common.token).update_counts()
         logging.info('Download Counter completed! (%s)', game_id)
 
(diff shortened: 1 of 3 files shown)


Clayell:
Hebaru privately brought up a point that I hadn't considered before:
If someone adopts a mod that has this modifier, and then they put that mod on CKAN, they would probably want the download count to include the mod that they're adopting, but not whatever parents the original mod wanted to block. A simple bool wouldn't allow this.

A potential workaround could be to allow the mod author to specify the level of recursion that they want to go, with the default being 3 and the "don't include parents" behavior being 0. Hebaru also brought up the issue that this would mean the person adopting the mod would need to remember to increment that counter by 1 when submitting their netkan. Perhaps a good way to remind people to do this would be to add a github action that checks if the new mod's parent (if it has one) has this modifier, and alerting the author in their netkan PR to increment it by one. (or does it for them)

I should fully test this behavior beforehand. There are existing forks of TIP (the mod that this was designed for), so I would probably make a fake netkan and make sure that the behavior (and github action) trigger correctly.

HebaruSan:
@Clayell, is the count in CKAN actually that important? What about just putting a download count badge in your README?

```markdown
<img src="https://img.shields.io/github/downloads/Clayell/TargetInterceptPlanner/total.svg" height="48px" style="height:48px;" />
```

Clayell:
If it wasn't important it wouldn't be on ckan...?

I also already have a download count in my readme.

